PR #81, opened 2024-06-10
Switch to SSZ Merkle Multiproofs

### LINKED ISSUE #55: Switch to SSZ Merkle Multiproofs
In the process of defining [Bolt specs](https://chainbound.github.io/bolt-docs/) we've decided to leverage SSZ Merkle Multiproofs (consensus specs [here](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) since they require less space and they're probably faster to compute as well, however the initial implementation consists of an array of single merkle proofs instead. 

This shouldn't be too much work but it is a sensible part and requires changes over the whole stack: mev-boost, relay and builder.

thedevbirb:
Closes #55.


merklefruit (approved):
very nice improvement overall, great job

PR #88, opened 2024-06-17
Integrate generic signer in sidecar

### LINKED ISSUE #63: Evaluate integrating `commit-boost` signing manager
## Context
Signing constraints & commitments has been left as something of a black box. If commit-boost is easy to integrate, we just use that and they manage all the security aspects of it. 

https://github.com/Commit-Boost/commit-boost-client/blob/main/crates/crypto/src/manager.rs

thedevbirb:
Closes #63. This PR introduces a generic BLS signer in the sidecar that could be obtained either by providing a private key or by integrating commit-boost


bolt-sidecar/bolt-sidecar/Cargo.lock
merklefruit, line 1: why is there a second bolt-sidecar directory with only this file inside?

bolt-sidecar/bolt.config.toml
thedevbirb, line 1: @merklefruit included this from the commit-boost devnet branch, do you if we actually need it for the tests? To me it doesn't seems so but I want to be sure

merklefruit: hmm, I don't know if it makes sense to keep, because the interface of that file will most likely change before commit-boost has a stable release. Let's keep it out for now IMO

merklefruit (approved):
Lgtm, after config file removed

Issue #24, opened 2024-05-14
Move proof verification logic to the sidecar


thedevbirb:
When the builder submits a bid with proofs of transaction inclusion, the relay propagates them and they're verified inside our `mev-boost` fork. However, since the sidecar is the entrypoint from the builder api and a proxy for `mev-boost` it makes sense to move proof verification there, and keep the `mev-boost` fork as light as possible.

merklefruit:
The reason we verify proofs in mev-boost is because we might otherwise pick a best_bid that has invalid proofs and send it to the sidecar. Closing this issue for now.

PR #89, opened 2024-06-17
feat(sidecar): build valid fallback payload

merklefruit:


bolt-sidecar/src/api/builder.rs
```diff
@@ -113,6 +113,8 @@ impl<T: ConstraintsApi, P: PayloadFetcher + Send + Sync> BuilderProxyServer<T, P
         {
             Ok(Ok(header)) => {
                 tracing::debug!(elapsed = ?start.elapsed(), "Returning signed builder bid: {:?}", header);
+                // TODO: verify proofs here. If they are invalid, we should fall back to locally built block
```
mempirate, line 116: We don't need to verify proofs again imo. We can assume mev-boost & relay have already done this. If mev-boost verification fails it will just return no bid

merklefruit: Makes sense, just thought it would be low cost, plus as a reminder for #24 which is still open

PR #98, opened 2024-07-01
feat(sidecar): consensus state validation

namn-grg:


bolt-sidecar/src/state/consensus.rs
```diff
@@ -40,6 +84,37 @@ impl ConsensusState {
...
+        Ok(())
+    }
+}
+
+/// Get the current timestamp.
+fn current_timestamp() -> u64 {
+    return SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
```
namn-grg, line 119: Just a util function, ideally should be in utils.rs (not sure of the file structure in rust)

namn-grg: NOTE - the precision of the deadline is in seconds. We might want to do it in mili seconds or smaller

merklefruit: Seconds is fine for now!

merklefruit (approved):
This is in the right direction! Happy to merge as is and iterate over it in further PRs to unlock some other work on this part.

PR #101, opened 2024-07-01
feat(sidecar): configurable `ValidatorIndex`

### LINKED ISSUE #57: Usage of `ValidatorIndex` inside Bolt Sidecar
Our current API definitions for constraints contain the `validatorIndex` of the validator that is opted-in to Bolt and has signed that specific preconfirmation request. 

This means that we have to deal with this when generating the constraints, which happens inside Bolt sidecar. 
Right now, this value is set to 0 always here: https://github.com/chainbound/bolt/pull/53/files#diff-709413e7867c3101338b843ff608b31b209deb676433b25160302d6819deafffR124

This issue is to keep track of this dangling data point and to brainstorm solutions. Some questions:

- should bolt sidecar work with >1 validators connected? if so, how do we figure out which validator to sign commitments with
- are there DDoS attacks that benefit from knowing the validatorIndex bound to some bolt sidecar?

namn-grg:
This PR adds CLI option to pass `validator_index` as comma-separated values i.e. `--validator-index "1,2,3,4"`

fixes #57

### DIFF
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -19,7 +19,6 @@ use bolt_sidecar::{
     BuilderProxyConfig, Config, MevBoostClient,
 };
 
-use beacon_api_client::ProposerDuty;
 use tokio::sync::mpsc;
 use tracing::info;
 
@@ -72,11 +71,14 @@ async fn main() -> eyre::Result<()> {
                 tracing::info!("Received commitment request: {:?}", event.request);
                 let request = event.request;
 
-                if let Err (e) = consensus_state.validate_request(&CommitmentRequest::Inclusion(request.clone())) {
-                    tracing::error!("Failed to validate request: {:?}", e);
-                    let _ = event.response.send(Err(ApiError::Custom(e.to_string())));
-                    continue;
-                }
+                let validator_index = match consensus_state.validate_request(&CommitmentRequest::Inclusion(request.clone()), validator_indexes) {
+                    Ok(index) => index,
+                    Err(e) => {
+                        tracing::error!("Failed to validate request: {:?}", e);
+                        let _ = event.response.send(Err(ApiError::Custom(e.to_string())));
+                        continue;
+                    }
+                };
 
                 if let Err(e) = execution_state
                     .try_commit(&CommitmentRequest::Inclusion(request.clone()))
@@ -91,8 +93,6 @@ async fn main() -> eyre::Result<()> {
                     tx_hash = %request.tx.tx_hash(),
                     "Validation against execution state passed"
                 );
-
-                let validator_index = find_validator_index_for_slot(validator_indexes, &consensus_state.get_epoch().proposer_duties, request.slot);
 
                 // parse the request into constraints and sign them with the sidecar signer
                 let message = ConstraintsMessage::build(validator_index, request.slot, request.clone());
@@ -141,50 +141,4 @@ async fn main() -> eyre::Result<()> {
     shutdown_tx.send(()).await.ok();
 
     Ok(())
-}
-
-/// Filters the proposer duties and returns the validator index for a given slot
-/// if it doesn't exists then returns 0 by default.
-pub fn find_validator_index_for_slot(
-    validator_indexes: &[u64],
-    proposer_duties: &[ProposerDuty],
-    slot: u64,
-) -> u64 {
-    for duty in proposer_duties {
-        if duty.slot == slot && validator_indexes.contains(&(duty.validator_index as u64)) {
-            return duty.validator_index as u64;
-        }
-    }
-    0
-}
-
-#[cfg(test)]
-mod tests {
-    use crate::find_validator_index_for_slot;
-    use beacon_api_client::ProposerDuty;
-
-    #[test]
-    fn test_filter_index() {
-        let validator_indexes = vec![11, 22, 33];
-        let proposer_duties = vec![
-            ProposerDuty {
-                public_key: Default::default(),
-                slot: 1,
-                validator_index: 11,
-            },
-            ProposerDuty {
-                public_key: Default::default(),
-                slot: 2,
-                validator_index: 22,
-            },
-            ProposerDuty {
-                public_key: Default::default(),
-                slot: 3,
-                validator_index: 33,
-            },
-        ];
-
-        let result = find_validator_index_for_slot(&validator_indexes, &proposer_duties, 2);
-        assert_eq!(result, 22);
-    }
 }
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -41,7 +41,7 @@ async fn main() -> eyre::Result<()> {
 
     let (api_events, mut api_events_rx) = mpsc::channel(1024);
     let shutdown_tx = json_rpc::start_server(&config, api_events).await?;
-    let consensus_state = ConsensusState::new(&config.beacon_api_url);
+    let consensus_state = ConsensusState::new(&config.beacon_api_url, &config.validator_indexes);
 
     let builder_proxy_config = BuilderProxyConfig {
         mevboost_url: config.mevboost_url,
@@ -50,8 +50,6 @@ async fn main() -> eyre::Result<()> {
 
     let (payload_tx, mut payload_rx) = mpsc::channel(16);
     let payload_fetcher = LocalPayloadFetcher::new(payload_tx);
-
-    let validator_indexes = &config.validator_indexes;
 
     tokio::spawn(async move {
         loop {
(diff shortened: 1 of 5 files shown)


bolt-sidecar/bin/sidecar.rs
```diff
@@ -89,9 +92,10 @@ async fn main() -> eyre::Result<()> {
                     "Validation against execution state passed"
                 );
 
+                let validator_index = find_validator_index_for_slot(&validator_indexes, &consensus_state.get_epoch().proposer_duties, request.slot);
```
merklefruit: We can keep this functionality inside ConsensusState's `validate_request` function's return data, like this:

```rust
pub fn validate_request(&self, request: &CommitmentRequest) -> Result<u64, ConsensusError> {
    ...
```

This way we can keep the executable cleaner. Here's how we could refactor line 75 then:

```rust
let validator_index = match consensus_state.validate_request(&CommitmentRequest::Inclusion(request.clone())) {
	Ok(index) => index,
	Err(e) => {    
		tracing::error!("Failed to validate request: {:?}", e);
		let _ = event.response.send(Err(ApiError::Custom(e.to_string())));
		continue;
	}
};
```

bolt-sidecar/bin/sidecar.rs
```diff
@@ -138,3 +142,49 @@ async fn main() -> eyre::Result<()> {
...
+    proposer_duties: &[ProposerDuty],
+    slot: u64,
+) -> u64 {
+    for duty in proposer_duties {
+        if duty.slot == slot && validator_indexes.contains(&(duty.validator_index as u64)) {
+            return duty.validator_index as u64;
+        }
+    }
+    0
+}
```
thedevbirb: A more idiomatic approach to do this (imo) is the following:
```rs
proposer_duties.iter().find(|duty| duty.slot == slot && ... )
```
There returns an `Option<ProposerDuty>` which is more flexible compared to returning `0` by default (maybe who uses this would like a different value). If the result is not `None` then we the caller can easily extract the validator index.
Wydt?

bolt-sidecar/bin/sidecar.rs
```diff
@@ -138,3 +142,49 @@ async fn main() -> eyre::Result<()> {
 
     Ok(())
 }
+
+/// Filters the proposer duties and returns the validator index for a given slot
+/// if it doesn't exists then returns 0 by default.
+pub fn find_validator_index_for_slot(
```
merklefruit: this function and tests can be moved to state/consensus.rs after the refactor above

bolt-sidecar/src/config.rs
```diff
@@ -27,6 +26,9 @@ pub struct Opts {
     /// Max commitments to accept per block
     #[clap(short = 'm', long)]
     pub(super) max_commitments: Option<usize>,
+    /// Validator indexes
+    #[clap(short = 'v', long)]
+    pub(super) validator_index: Option<String>,
```
thedevbirb: With `clap` you can use a specific parser using
`#[clap(short = 'v', long, value_parser = "...")]`
or something a long these lines!
Last nit, this should be called `validator_indexes` as well!

bolt-sidecar/src/config.rs
```diff
@@ -1,8 +1,7 @@
+use crate::crypto::bls::random_bls_secret;
 use blst::min_pk::SecretKey;
 use clap::{ArgGroup, Args, Parser};
 
-use crate::crypto::bls::random_bls_secret;
-
```
merklefruit: nit: imports are usually ordered automatically by rustfmt, with the standard:

- std library imports
- external crates
- crate:: and self:: imports

each category separated by a new line.
This isn't a strict requirement at all, and you'll often find different orders in some rust files,
but I thought it would be good if we're aligned on a standard :)

bolt-sidecar/src/config.rs
```diff
@@ -82,6 +86,7 @@ impl Default for Config {
             private_key: Some(random_bls_secret()),
             mevboost_proxy_port: 18551,
             limits: Limits::default(),
+            validator_indexes: vec![0], // Default to 0 index
```
merklefruit: I think it makes more sense to initialize this to an empty array 

```suggestion
            validator_indexes: Vec::new()
```

merklefruit (changes requested):
Looks good! Left a few nits. Just one note: have you been able to test the CLI option?
I don't think that `--validator-index 1, 3, 4` would work because of the spaces after each comma, but I'm not sure. 

Also, it would be nice to support specifying an array with the [2,5] notation specified in the issue. One way to do this could be to check if the flag value starts with `[` in the parse_validator_indexes function, but maybe there are other ways too

bolt-sidecar/src/config.rs
namn-grg: ok!

bolt-sidecar/src/config.rs
thedevbirb: Actually nvm about `value_parser`, we have a `try_from` for the config with parsing logic there.

namn-grg: > Actually nvm about `value_parser`, we have a `try_from` for the config with parsing logic there.

I didn't get this. I tried doing it with `value_parser` but faced errors. Now how should I do it again? or is it resolved?

bolt-sidecar/bin/sidecar.rs
namn-grg: Great idea! Nit: we'll also need to pass the validator_indexes in the argument

merklefruit: @namn-grg We can initialize the `ConsensusState` with the `validator_indexes` values, such that we don't have to pass the argument each time.

bolt-sidecar/bin/sidecar.rs
namn-grg: yes I agree, I have updated the fn now in a similar way

bolt-sidecar/bin/sidecar.rs
namn-grg: Good point! made changes

bolt-sidecar/src/config.rs
namn-grg: Got it, I'll remember this

namn-grg:
> Looks good! Left a few nits. Just one note: have you been able to test the CLI option? I don't think that `--validator-index 1, 3, 4` would work because of the spaces after each comma, but I'm not sure.
> 
> Also, it would be nice to support specifying an array with the [2,5] notation specified in the issue. One way to do this could be to check if the flag value starts with `[` in the parse_validator_indexes function, but maybe there are other ways too

@merklefruit as discussed since the brackets are an issue, I have changed it to be compatible with quotation marks i.e. `--validator-index "1,2,3,4"`

bolt-sidecar/src/config.rs
```diff
@@ -27,6 +27,9 @@ pub struct Opts {
     /// Max commitments to accept per block
     #[clap(short = 'm', long)]
     pub(super) max_commitments: Option<usize>,
+    /// Validator indexes
+    #[clap(short = 'v', long)]
+    pub(super) validator_indexes: Option<String>,
```
merklefruit: Clap has a feature called `value_parser` (docs [here](https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes) and [here](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser)) that's really cool.
I think that with this we could do without the `parse_validator_indexes` function

```suggestion
    /// Validator indexes
    #[clap(short = 'v', long, value_parser, num_args = 0.., value_delimiter = ',')]
    pub(super) validator_indexes: Vec<String>,
```

namn-grg: ah good find again! I should have looked at the docs. Nit - I made the type as `Vec<u64>` (not string) as it can be parsed directly now

merklefruit: Awesome!

merklefruit (approved):
Great job with this one :)

bolt-sidecar/src/state/consensus.rs
```diff
@@ -118,3 +133,50 @@ fn current_timestamp() -> u64 {
         .unwrap()
         .as_secs()
 }
+
+/// Filters the proposer duties and returns the validator index for a given slot
+/// if it doesn't exists then returns 0 by default.
+pub fn find_validator_index_for_slot(
```
mempirate: This should be a method on `ConsensusState`, which already has `validator_indexes` and `proposer_duties` as fields

bolt-sidecar/src/config.rs
```diff
@@ -27,6 +27,9 @@ pub struct Opts {
     /// Max commitments to accept per block
     #[clap(short = 'm', long)]
     pub(super) max_commitments: Option<usize>,
+    /// Validator indexes
+    #[clap(short = 'v', long, value_parser, num_args = 0.., value_delimiter = ',')]
```
thedevbirb: Shouldn't `num_args` be `1..` instead of `0..`? In order for Bolt to function properly you must specify at least one validator index

mempirate (changes requested):
Looking good, small nit and updated some stuff as well

thedevbirb (approved):
Last nit, great job!

bolt-sidecar/src/state/consensus.rs
namn-grg: good catch, done

bolt-sidecar/src/config.rs
namn-grg: good catch, done

PR #104, opened 2024-07-03
fix: self-building payloads in kurtosis devnet

merklefruit:
- [x] fixed: signature error on `get_header` response with fallback payloads
- [x] fixed: added fork_version to compute builder_domain for signing builder-api messages
- [x] fixed: wrong return type for get_payload (due to serde enum tag versioning)
- [x] fixed: SSZ transactions_root and withdrawals_root in consensus execution payload header instead of MPT

### DIFF
--- a/bolt-sidecar/src/builder/mod.rs
+++ b/bolt-sidecar/src/builder/mod.rs
@@ -115,7 +115,7 @@ impl LocalBuilder {
         let payload_and_blobs = PayloadAndBlobs {
             execution_payload: eth_payload,
             // TODO: add included blobs here
-            blobs_bundle: None,
+            blobs_bundle: Default::default(),
         };
 
         // 2. create a signed builder bid with the sealed block header we just created
--- a/bolt-sidecar/src/builder/mod.rs
+++ b/bolt-sidecar/src/builder/mod.rs
@@ -1,19 +1,27 @@
-use alloy_primitives::{Address, U256};
+use alloy_primitives::U256;
+use blst::min_pk::SecretKey;
 use ethereum_consensus::{
-    crypto::SecretKey as BlsSecretKey,
-    ssz::prelude::{HashTreeRoot, List, MerkleizationError},
+    crypto::PublicKey,
+    ssz::prelude::{List, MerkleizationError},
 };
 use payload_builder::FallbackPayloadBuilder;
 use reth_primitives::{SealedHeader, TransactionSigned};
+use signature::sign_builder_message;
 
-use crate::primitives::{
-    BuilderBid, GetPayloadResponse, PayloadAndBid, PayloadAndBlobs, SignedBuilderBid,
+use crate::{
+    primitives::{
+        BuilderBid, GetPayloadResponse, PayloadAndBid, PayloadAndBlobs, SignedBuilderBid,
+    },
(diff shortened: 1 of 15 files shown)


bolt-sidecar/src/builder/mod.rs
```diff
@@ -117,13 +115,20 @@ impl LocalBuilder {
         let payload_and_blobs = PayloadAndBlobs {
             execution_payload: eth_payload,
             // TODO: add included blobs here
-            blobs_bundle: None,
+            blobs_bundle: Default::default(),
```
merklefruit, line 118: we should add an issue for this: support blobs across the self building pipeline

bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -113,19 +106,21 @@ impl FallbackPayloadBuilder {
...
-        let withdrawals = if withdrawals.is_empty() {
-            None
-        } else {
-            Some(withdrawals)
-        };
+        // let withdrawals = if withdrawals.is_empty() {
+        //     None
+        // } else {
+        //     Some(withdrawals)
+        // };
```
thedevbirb: We can drop this comment!

bolt-sidecar/src/config.rs
```diff
@@ -63,9 +60,85 @@ pub struct Opts {
...
+
+    /// Get the slot time for the given chain in seconds.
+    pub fn slot_time(&self) -> u64 {
+        match self {
+            Chain::Mainnet => 12,
+            Chain::Holesky => 12,
+            Chain::Kurtosis => 2,
+            Chain::Helder => 12,
+        }
+    }
```
thedevbirb, line 111: I'd still keep the slot time as a separate CLI arg, with default 12, in case we want to run the devnet with a more realistic slot time. We can do it in a separate PR

bolt-sidecar/src/config.rs
```diff
@@ -63,9 +60,85 @@ pub struct Opts {
...
+        match self {
+            Chain::Mainnet => {
+                b256!("00000001f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a9").0
+            }
+            Chain::Holesky => {
+                b256!("000000015b83a23759c560b2d0c64576e1dcfc34ea94c4988f3e0d9f77f05387").0
+            }
+            Chain::Kurtosis => {
+                // TODO: verify this
+                // b256!("00000001f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a9").0
```
thedevbirb: We can drop this comments now!

thedevbirb (approved):
LFG

bolt-sidecar/src/config.rs
merklefruit: Yeah we should refactor the `Chain` configs to another file perhaps

bolt-sidecar/src/builder/mod.rs
merklefruit: #105

bolt-sidecar/src/config.rs
merklefruit: #106

PR #117, opened 2024-07-04
Submit constraints only after commitments deadline + block template refactor

### LINKED ISSUE #102: Sidecar: check user balance and nonce before sending commitments
In the context of the recent proposal of account abstraction it may be possible that a transaction changes the balance of an EOA without increasing its nonce, since it can be originated by another entity.
As such, in the sidecar we should:
- wait for the right slot before sending the commitments and
- check both balance and nonce according to the current state

### LINKED ISSUE #109: Submit constraints only after the commitment deadline
As we can see in the main sidecar loop here
https://github.com/chainbound/bolt/blob/4f87242f70ec11ccbb01e32852703bcf17fc6d6a/bolt-sidecar/bin/sidecar.rs#L68-L116
constraints are submitted immediately to MEV-Boost after quick checks on protocol rules (nonce, balance, base fee). In turn, this constraints are then forwarded to relays and pulled from builders. 

This is actually not desirable behaviour because a pre-confirmation could be invalidated if, in the slot where the pre-confirmation will be executed, the user has an higher nonce or an insufficient balance. 
In such case the proposer won't be slashed because it can prove on-chain what happened, however the builder won't be able to produce a bid with all the proofs which will be discarded by MEV-Boost, leading to a sub-optimal locally built block.

In order to solve this, pre-confirmations should be cached by their slot and when a new head comes we should check again their validity. Lastly, constraints should be submitted only after the commit deadline for a given slot is crossed.

### LINKED ISSUE #125: Sidecar: invalidate past state
Block templates should be cleared once a new head for that slot has been received from the CL.
Furthermore, tracked nonces should be only reset if a preconfirmed transaction was successfully included in a block.

https://github.com/chainbound/bolt/blob/6091199cd8ad7c959d2cbe79852f9ec16d7a2c64/bolt-sidecar/src/state/execution.rs#L201

thedevbirb:
This PR refactors the block template by using the signed constraints list instead of the transactions signed. This allows a more granular control over the current state of the pre-confirmations accepted, by dropping an entire set of constraints in case some of its pre-confirmed transactions have been invalidated.

- closes #109 
- closes #102 
- closes #125 

## Summary by CodeRabbit

- **New Features**
  - Introduced pre-confirmation spammer for Helder testnet with environment variable configuration.
  - Added functions for generating random Ethereum transactions and fetching the current slot using a Beacon API client.

- **Enhancements**
  - Improved commitment validation and signing processes.
  - Integrated constraints into the execution state with enhanced retry logic for constraint submissions.
  - Optimized gas-related adjustments and transaction handling across various modules.

- **Bug Fixes**
  - Refined error handling for hexadecimal parsing and decoding.

- **Chores**
  - Added `.env` to `.gitignore` to exclude sensitive information from version control.
  - Updated dependencies in `Cargo.toml` to reflect new project requirements.

- **Documentation**
  - Updated `README.md` to provide instructions for running the new pre-confirmation spammer.
  - Added `rustfmt.toml` for consistent code formatting.


bolt-sidecar/bin/sidecar.rs
```diff
@@ -77,43 +78,34 @@ async fn main() -> eyre::Result<()> {
...
-                    .await
+                let sender = match execution_state
+                    .check_commitment_validity(&request)
+                .await
                 {
-                    tracing::error!("Failed to commit request: {:?}", e);
-                    let _ = response_tx.send(Err(ApiError::Custom(e.to_string())));
-                    continue;
-                }
+                    Ok(sender) => { sender },
```
merklefruit: how are the {} not removed by rustfmt? first time I see it like this hahah

bolt-sidecar/src/builder/template.rs
```diff
@@ -35,82 +38,145 @@ impl BlockTemplate {
...
+                    *balance += diff.balance;
+                })
+                .or_insert((diff.transaction_count, diff.balance));
+        });
+
+        self.signed_constraints_list.push(signed_constraints);
+    }
 
-        self.transactions.push(transaction);
+    /// Returns all a clone of all transactions from the signed constraints list
```
merklefruit:
```suggestion
    /// Returns a copy of all transactions from the signed constraints list
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -79,20 +82,26 @@ pub struct ExecutionState<C> {
...
+        client: C,
+        max_commitments_per_slot: NonZero<usize>,
+    ) -> Result<Self, TransportError> {
         Ok(Self {
             basefee: client.get_basefee(None).await?,
             block_number: client.get_head().await?,
             slot: 0,
             account_states: HashMap::new(),
             block_templates: HashMap::new(),
+            max_commitments_per_slot,
```
namn-grg: Just a soft note, I skipped doing like this for adding chain_id validation (see https://github.com/chainbound/bolt/pull/126#discussion_r1670270048 ) because looked unclean to me but if we have to do it here then we can consider it there also.

bolt-sidecar/src/builder/template.rs
```diff
@@ -24,8 +27,8 @@ use crate::{common::max_transaction_cost, primitives::AccountState};
 pub struct BlockTemplate {
     /// The state diffs per address given the list of commitments.
     state_diff: StateDiff,
-    /// The list of transactions in the block template.
-    pub transactions: Vec<TransactionSigned>,
+    /// The signed constraints associated to the block
+    pub signed_constraints_list: Vec<SignedConstraints>,
```
merklefruit, line 35: this is awesome

bolt-sidecar/src/builder/template.rs
```diff
@@ -35,82 +38,145 @@ impl BlockTemplate {
...
-            acc
+        self.signed_constraints_list.iter().fold(0, |acc, sc| {
+            acc + sc
+                .message
+                .constraints
+                .iter()
+                .filter(|c| c.tx_decoded.tx_type() == TxType::Eip4844)
+                .count()
         })
     }
```
merklefruit: this used to count the number of blobs, while now it counts the number of blob transactions. 
Since each transaction can carry up to 6 blobs, we should either call this function `blob_tx_count` or reimplement the count of versioned hashes to replicate the previous behaviour.

bolt-sidecar/src/builder/template.rs
```diff
@@ -35,82 +38,145 @@ impl BlockTemplate {
...
+            .iter()
+            .flat_map(|c| c.1.clone())
+            .fold((U256::ZERO, u64::MAX), |mut acc, c| {
+                let nonce = c.tx_decoded.nonce();
+                if nonce < acc.1 {
+                    acc.1 = nonce;
+                }
+                acc.0 += max_transaction_cost(&c.tx_decoded);
+                acc
+            });
```
merklefruit: we can avoid cloning + use named variables instead of the tuple here, wdyt?

```suggestion
            .flat_map(|(_, c)| c.as_slice())
            .fold((U256::ZERO, u64::MAX), |(total_cost, min_nonce), c| {
                (
                    total_cost + max_transaction_cost(&c.tx_decoded),
                    min_nonce.min(c.tx_decoded.nonce()),
                )
            });
```

bolt-sidecar/src/builder/template.rs
```diff
@@ -35,82 +38,145 @@ impl BlockTemplate {
...
+
+        if state.balance < max_total_cost || state.transaction_count > min_nonce {
+            // NOTE: We drop all the signed constraints containing such pre-confirmations
+            // since at least one of them has been invalidated.
+            tracing::warn!(
+                %address,
+                "Removing all signed constraints which contain such address pre-confirmations due to conflict with account state",
+            );
+            indexes = constraints_with_address.iter().map(|c| c.0).collect();
```
merklefruit:
```suggestion
            indexes = constraints_with_address.iter().map(|(idx, _)| *idx).collect();
```

bolt-sidecar/src/builder/template.rs
```diff
@@ -35,82 +38,145 @@ impl BlockTemplate {
...
+                acc.0 += max_transaction_cost(&c.tx_decoded);
+                acc
+            });
+
+        if state.balance < max_total_cost || state.transaction_count > min_nonce {
+            // NOTE: We drop all the signed constraints containing such pre-confirmations
+            // since at least one of them has been invalidated.
+            tracing::warn!(
+                %address,
+                "Removing all signed constraints which contain such address pre-confirmations due to conflict with account state",
```
merklefruit:
```suggestion
                %address,
                "Removing all signed constraints by {address:?} due to conflicts with account state",
```

bolt-sidecar/src/state/execution.rs
merklefruit: but in this case,  we can't fetch `max_commitments_per_slot` from the client, unlike in your PR where we use the client to read the chain_id. I might have misunderstood what you mean though

bolt-sidecar/src/state/execution.rs
```diff
@@ -115,9 +124,21 @@ impl<C: StateFetcher> ExecutionState<C> {
...
+    ) -> Result<Address, ValidationError> {
         let CommitmentRequest::Inclusion(req) = request;
 
+        // Check if there is room for more commitments
+        if let Some(template) = self.get_block_template(req.slot) {
+            if template.transactions_len() >= self.max_commitments_per_slot.get() {
+                return Err(ValidationError::MaxCommitmentsReachedForSlot(
+                    self.max_commitments_per_slot.get(),
+                ));
+            }
```
merklefruit: From the `MaxCommitmentsReachedForSlot` error description below, it seems like we should use `req.slot` here instead, or am I misinterpreting it?

```rust
    #[error("Max commitments reached for slot {0}")]
    MaxCommitmentsReachedForSlot(usize),
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -115,9 +124,21 @@ impl<C: StateFetcher> ExecutionState<C> {
     /// If the commitment is valid, it will be added to the block template and its account state
     /// will be cached. If this is succesful, any callers can be sure that the commitment is valid
     /// and SHOULD sign it and respond to the requester.
-    pub async fn try_commit(&mut self, request: &CommitmentRequest) -> Result<(), ValidationError> {
+    pub async fn check_commitment_validity(
```
merklefruit: I like that you cached the ec-recovery operation, great idea

merklefruit (changes requested):
Great job overall. One general nit: in Milan we had agreed to always use the term `preconfirmation` which we use across other Bolt components, so it would be great if we kept consistent with it instead of having both it and `pre-confirmation` in the codebase

bolt-sidecar/bin/sidecar.rs
mempirate: Because it's inside a macro (`select!`) 👎

bolt-sidecar/src/builder/template.rs
mempirate: Good catch, changed in upstack

bolt-sidecar/src/builder/template.rs
thedevbirb: Thanks for the cleanup!

bolt-sidecar/src/builder/template.rs
thedevbirb: Updated in the upstack

bolt-sidecar/src/builder/template.rs
thedevbirb: Changed in upstack

bolt-sidecar/src/state/execution.rs
thedevbirb: Good catch, adding both the slot and the maximum number of constraints reached

PR #119, opened 2024-07-05
chore(sidecar): add parsing validator indexes as ranges

merklefruit:
This PR adds the ability to parse validator index ranges from an array of comma-separated ranges.
ValidatorIndex also becomes a separate struct to allow for future parsing methods too, including from a keystore directory which would be handy for validators

example of valid values:

```
--validator-indexes 1,2,3,4,5
--validator-indexes 1..3,5..7
--validator-indexes 1,2,3,67..97,194,529..541
```


namn-grg (approved):
LGTM

thedevbirb (approved):

PR #127, opened 2024-07-12
fix(sidecar): deserialize get_payload response from mev-boost

merklefruit:
The manual deserialize implementation for `GetPayloadResponse` [here](https://github.com/chainbound/bolt/blob/721ffd792f52487a40b804fa3caab016dd035510/bolt-sidecar/src/primitives/mod.rs#L194) was incorrect. Fixed by removing it. 

In an intermediate debug build, I saved a raw get_payload_response to a file and added it to the test_util client module, such that deserialization can be properly tested in the codebase from a real value. 

This PR bumps the sidecar image version to `v0.1.1-alpha.3-rc4`


bolt-sidecar/src/client/test_util/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[test]
+fn test_decode_get_payload_response() {
+    let stringified =
+        std::fs::read_to_string("./src/client/test_util/deneb_get_payload_response.json")
+            .expect("failed to read get payload response file");
+
+    let parsed_response: GetPayloadResponse =
+        serde_json::from_str(&stringified).expect("failed to parse get payload response");
+}
```
merklefruit, line 110: added this test using a real value read from file

thedevbirb (approved):
Great!

PR #128, opened 2024-07-12
Sidecar: preliminary blob validation support

mempirate:
Adds preliminary blob validation & fixes some tests.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method `get_blob_basefee` in the client to fetch blob base fees.
  - Added `random()` method to create random BLS keys in the `Signer` implementation.

- **Improvements**
  - Enhanced handling and validation of commitment requests for better error handling.
  - Improved error handling for missing RPC URLs in environment variables.
  - Refined transaction and constraint processing in `BlockTemplate`.

- **Refactor**
  - Updated method and function names to improve code clarity.
  - Replaced `TransactionSigned` with `PooledTransactionsElement` in several modules for consistency.

### DIFF
--- a/bolt-sidecar/src/builder/state_root.rs
+++ b/bolt-sidecar/src/builder/state_root.rs
@@ -10,7 +10,9 @@ mod tests {
     use partial_mpt::StateTrie;
     use reqwest::Url;
 
-    use crate::{builder::CallTraceManager, client::rpc::RpcClient};
+    use crate::{
+        builder::CallTraceManager, client::rpc::RpcClient, test_util::try_get_execution_api_url,
+    };
 
     #[ignore]
     #[tokio::test]
@@ -18,14 +20,14 @@ mod tests {
         dotenvy::dotenv().ok();
         let _ = tracing_subscriber::fmt::try_init();
 
-        let Some(rpc_url) = std::env::var("RPC_URL").ok() else {
-            tracing::warn!("RPC_URL not found in environment variables, skipping test");
+        let Some(rpc_url) = try_get_execution_api_url().await else {
+            tracing::warn!("EL_RPC not reachable, skipping test");
             return Ok(());
         };
 
         tracing::info!("Starting test_trace_call");
 
-        let rpc_url = Url::parse(&rpc_url).unwrap();
+        let rpc_url = Url::parse(rpc_url).unwrap();
         let client = RpcClient::new(rpc_url.clone());
 
         let (call_trace_manager, call_trace_handler) = CallTraceManager::new(rpc_url);
--- a/bolt-sidecar/src/builder/state_root.rs
+++ b/bolt-sidecar/src/builder/state_root.rs
@@ -12,6 +12,7 @@ mod tests {
 
     use crate::{builder::CallTraceManager, client::rpc::RpcClient};
 
+    #[ignore]
     #[tokio::test]
     async fn test_trace_call() -> eyre::Result<()> {
         dotenvy::dotenv().ok();
--- a/bolt-sidecar/src/builder/state_root.rs
+++ b/bolt-sidecar/src/builder/state_root.rs
@@ -17,9 +17,13 @@ mod tests {
         dotenvy::dotenv().ok();
         let _ = tracing_subscriber::fmt::try_init();
 
+        let Some(rpc_url) = std::env::var("RPC_URL").ok() else {
+            tracing::warn!("RPC_URL not found in environment variables, skipping test");
+            return Ok(());
+        };
+
         tracing::info!("Starting test_trace_call");
 
-        let rpc_url = std::env::var("RPC_URL").expect("RPC_URL must be set");
         let rpc_url = Url::parse(&rpc_url).unwrap();
         let client = RpcClient::new(rpc_url.clone());
 
--- a/bolt-sidecar/src/builder/state_root.rs
+++ b/bolt-sidecar/src/builder/state_root.rs
@@ -15,7 +15,7 @@ mod tests {
     #[tokio::test]
     async fn test_trace_call() -> eyre::Result<()> {
         dotenvy::dotenv().ok();
-        tracing_subscriber::fmt::init();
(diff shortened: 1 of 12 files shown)


mempirate:
@coderabbitai review

bolt-sidecar/src/builder/template.rs
```diff
@@ -28,7 +28,7 @@ pub struct BlockTemplate {
     /// The state diffs per address given the list of commitments.
     state_diff: StateDiff,
     /// The signed constraints associated to the block
-    pub signed_constraints_list: Vec<SignedConstraints>,
+    pub constraints: Vec<SignedConstraints>,
```
thedevbirb: I don't want to be pedantic but I think keeping the field named as `signed_constraints_list` is actually clearer, since the type `SignedConstraints` contains a `constraints` property as well which could lead to some ambiguity.

bolt-sidecar/src/builder/template.rs
```diff
@@ -37,101 +37,82 @@ impl BlockTemplate {
...
+                .entry(constraint.sender)
+                .and_modify(|(nonce, balance)| {
+                    *nonce += 1;
+                    *balance += max_cost;
+                })
+                .or_insert((1, max_cost));
+        }
+
+        self.constraints.push(constraints);
+    }
```
thedevbirb: This is much cleaner I don't know why I came up with that first implementation with an intermediate step lol, same for `remove_constraints_at_index`

bolt-sidecar/src/client/rpc.rs
```diff
@@ -45,6 +45,18 @@ impl RpcClient {
...
+    pub async fn get_blob_basefee(&self, block_number: Option<u64>) -> TransportResult<u128> {
+        let tag = block_number.map_or(BlockNumberOrTag::Latest, BlockNumberOrTag::Number);
+
+        let fee_history: FeeHistory = self
+            .0
+            .request("eth_feeHistory", (U64::from(1), tag, &[] as &[f64]))
+            .await?;
+
+        Ok(fee_history.latest_block_blob_base_fee().unwrap_or(0))
+    }
```
thedevbirb:
```suggestion
    /// Get the blob basefee of the latest block.
    ///
    /// Reference: https://github.com/ethereum/execution-apis/blob/main/src/eth/fee_market.yaml
    pub async fn get_blob_basefee(&self, block_number: Option<u64>) -> TransportResult<u128> {
        let block_count = U64::from(1);
        let tag = block_number.map_or(BlockNumberOrTag::Latest, BlockNumberOrTag::Number);
        let reward_percentiles: Vec<f64> = vec![];

        let fee_history: FeeHistory = self
            .0
            .request("eth_feeHistory", (block_count, tag, &reward_percentiles))
            .await?;

        Ok(fee_history.latest_block_blob_base_fee().unwrap_or(0))
    }
```
I think it is clearer to read it this way even if we actually don't use all the parameters

bolt-sidecar/src/builder/state_root.rs
```diff
@@ -15,11 +15,15 @@ mod tests {
     #[tokio::test]
     async fn test_trace_call() -> eyre::Result<()> {
         dotenvy::dotenv().ok();
-        tracing_subscriber::fmt::init();
+        let _ = tracing_subscriber::fmt::try_init();
+
+        let Some(rpc_url) = std::env::var("RPC_URL").ok() else {
+            tracing::warn!("RPC_URL not found in environment variables, skipping test");
+            return Ok(());
+        };
```
merklefruit: we have `test_util::try_get_execution_api_url()` which will try to ping the EL from our Helder tailnet and use it.
Env vars are also a great option, just wanted to point this out

bolt-sidecar/src/crypto/bls.rs
```diff
@@ -62,6 +62,13 @@ impl Signer {
         Self { key }
     }
 
+    /// Create a signer with a random BLS key.
+    pub fn random() -> Self {
+        Self {
+            key: random_bls_secret(),
+        }
+    }
```
merklefruit, line 70: awesome

bolt-sidecar/src/primitives/constraint.rs
```diff
@@ -85,32 +85,25 @@ impl SignableBLS for ConstraintsMessage {
...
-    pub(crate) tx_decoded: TransactionSigned,
+    /// The transaction to be included in the block
+    pub(crate) transaction: PooledTransactionsElement,
     /// The ec-recovered address of the transaction sender for internal use
     #[serde(skip)]
     pub(crate) sender: Address,
 }
 
 impl Constraint {
     /// Builds a constraint from an inclusion request and an optional index
```
merklefruit:
```suggestion
    /// Builds a constraint from a transaction, with an optional index
```

thedevbirb (approved):
Great job! Just some nits

bolt-sidecar/src/state/mod.rs
```diff
@@ -314,7 +335,73 @@ mod tests {
...
+            state
+                .block_templates()
+                .get(&target_slot)
+                .unwrap()
+                .transactions_len()
+                == 1
+        );
+
+        // Update the head, which should invalidate the transaction due to a nonce conflict
+        state.update_head(None, target_slot).await.unwrap();
```
merklefruit, line 403: great, what other edge cases should be tested in terms of constraint invalidation?

merklefruit (approved):
LGTM

bolt-sidecar/src/builder/template.rs
merklefruit: let's do signed_constraints as the plural (s) already hints to a list? wdyt?

bolt-sidecar/src/state/mod.rs
thedevbirb: Just the protocol rules for validity, therefore insufficient balance according to what's needed: the tx value + base fee per gas (we're probably ignoring the priority fee to the proposers until we've figured out pricing)

PR #130, opened 2024-07-15
chore: add commitment validity tests

merklefruit:
- refactored execution tests to use a streamlined test_util module for signing and creating commitment requests
- added the `has_code` property to the AccountState fetched during commitment request validation.
- added other minor commitment validity checks such as gas_limit and tx_size

Having the account code check prevents accepting preconfirmation requests from smart contracts addresses. 
Not entirely sure how this would be possible in the real world, but [EL clients](https://github.com/paradigmxyz/reth/blob/9aa44e1a90b262c472b14cd4df53264c649befc2/crates/transaction-pool/src/validate/eth.rs#L153) are also performing this check.


bolt-sidecar/src/state/execution.rs
```diff
@@ -100,6 +114,27 @@ pub struct ExecutionState<C> {
...
+
+impl Default for Constants {
+    fn default() -> Self {
+        Self {
+            block_gas_limit: 30_000_000,
+            max_tx_input_bytes: 4 * 32 * 1024, // 128kb
+            max_init_code_byte_size: 2 * 24576,
+        }
+    }
 }
```
thedevbirb: Why are you putting these constants in a struct field compared to being just constants in the code?

thedevbirb (commented):
Great job, just a q

bolt-sidecar/src/state/execution.rs
merklefruit: Because these should be configurable in theory, but I'm not fully convinced if this is useful now.
What makes more sense in your opinion?

thedevbirb: Why should they be configurable in theory? Just curious, I thought this values could change only with an hard-fork

merklefruit: these values are chosen by the EL, and can be configured via CLI, without hardforks

thedevbirb (approved):

PR #133, opened 2024-07-16
wip(sidecar): low-level `SidecarDriver`

### LINKED ISSUE #99: Sidecar StateDriver
The responsibilities of the sidecar main loop are starting to grow as we develop the fallback block building and consensus state container. Right now the main loop is exposed in the executable `bin/sidecar.rs`, but ideally that file should just contain initialization code and not business logic. 

To clean this up, we will create a `StateDriver` component with the following responsibilities:

1. subscribe to the latest head updates from the `HeadTracker`
2. `run` function with an infinite loop:
	- await for a new head event to be emitted
	- try to update the execution_state with its `update_head()` method
	- try to update the consensus_state with its `update_head()` method
	- in the future: handle reorgs as well

merklefruit:
This PR introduces a driver task for the main loop to replace the `tokio::select!` usage in our main binary.
This is an experiment and probably not worth to merge until we have a more stable release of the sidecar.

- closes #99 

To make the remaining handlers sync, we essentially need to process these 3 events synchronously:
1. update head (execution)
2. update head (consensus)
3. validate request (execution): this is async because we may need to fetch some missing state from the node via API

The best way to do this is by turning execution and consensus state containers into actors and control them from the driver by their handles.

### DIFF
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -104,7 +104,7 @@ async fn main() -> eyre::Result<()> {
                 let res = serde_json::to_value(signed_constraints).map_err(Into::into);
                 let _ = response_tx.send(res).ok();
             },
-            Ok(HeadEvent { slot, .. }) = head_tracker.next_head() => {
+            Some(HeadEvent { slot, .. }) = head_tracker.next_head() => {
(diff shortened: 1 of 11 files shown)


merklefruit:
closed in favor of #181

PR #135, opened 2024-07-16
Fix devnet with constraints deadline

thedevbirb:
- fixed a bug in the builder constraint cache timing: since we are now sending commitments towards the end of the slot, we cannot rely on fetching them when we start building for a specific slot. Instead, we keep the constraints cache downstream and constantly fetch from it to get the latest values each time a new worker is run.
- fixed an issue where the sidecar would throw a "payload header equivocation" error due to the local_payload cache not being cleared when getting a signed header back from the CL.


bolt-sidecar/src/api/builder.rs
```diff
@@ -127,6 +125,11 @@ where
             Ok(res) => match res {
                 Err(builder_err) => builder_err,
                 Ok(header) => {
+                    // Clear the local payload cache if we have a successful response
+                    // By definition of `server.local_payload`, this will be `Some` IFF we have signed a local header
+                    let mut local_payload = server.local_payload.lock();
+                    *local_payload = None;
```
merklefruit, line 131: since local_payload doesn't have a slot number, it was possible to get a stale block from this and result in a conflicting block hash in get_payload.

PR #137, opened 2024-07-16
fix(sidecar): validate state with existing diffs

### LINKED ISSUE #136: State diffs are not used for validating new preconfirmation requests
The validation of incoming preconfirmation requests is currently flawed because the nonce and balance checks are performed against the current account state (from the latest known head) but without considering prior preconfirmations (i.e., the state diffs). To solve this, we should check against both the account state and the state diffs.

Example:
We're at slot $N$ and I make two preconfirmation requests for slot $N+2$. The validation of the second preconfirmation is successful because it is not considering the temporary state difference generated from the first request.

thedevbirb:
Closes #136


bolt-sidecar/src/json_rpc/api.rs
```diff
@@ -113,7 +113,9 @@ impl CommitmentsRpc for JsonRpcApi {
 
         info!(?request, "received inclusion commitment request");
 
-        let tx_sender = request.tx.recover_signer().ok_or(ApiError::Custom(
+        // NOTE: request.sender is skipped from deserialization and initialized as Address::ZERO
+        // by the default Deserialization. It must be set here.
+        request.sender = request.tx.recover_signer().ok_or(ApiError::Custom(
             "failed to recover signer from transaction".to_string(),
         ))?;
```
merklefruit, line 120: any better ways to do this that don't involve adding an `Option<Address>` in `InclusionRequest`?

thedevbirb: We have thought about it but ultimately this leads to create a different type with just the ec-recovered sender, which means a lot of verbosity and duplicated types. 
As such, we came to the conclusion that this little "gotcha" is better for now

PR #150, opened 2024-07-19
feat(sidecar): add max_committed_gas_per_slot to sidecar Limits

### LINKED ISSUE #144: feat(sidecar): add `max_committed_gas_per_slot` to sidecar `Limits`
Bolt sidecar operators should have a way to limit both the number and the gas amount used for commitments in a single slot.
For this, we already have the `Limits` config struct that can be extended.

These limits are enforced in the execution state validation pipeline here:

https://github.com/chainbound/bolt/blob/7366aef7308ddaf9377b94d7af2eadf1746fe2ce/bolt-sidecar/src/state/execution.rs#L204-L210

it would also be ideal to place `Limits` inside the `ExecutionState` itself instead of copying the individual fields like this:

https://github.com/chainbound/bolt/blob/7366aef7308ddaf9377b94d7af2eadf1746fe2ce/bolt-sidecar/src/state/execution.rs#L115

namn-grg:
fixes #144 

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced gas management parameters for improved performance and resource management.
	- Added a method to calculate total committed gas in the BlockTemplate for better performance metrics.

- **Bug Fixes**
	- Improved error handling with a new validation error type for gas commitment limits.

- **Documentation**
	- Updated configuration documentation to reflect new parameters related to gas management.

### DIFF
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -28,12 +28,7 @@ async fn main() -> eyre::Result<()> {
     let signer = Signer::new(config.private_key.clone().unwrap());
 
     let state_client = StateClient::new(config.execution_api_url.clone());
-    let mut execution_state = ExecutionState::new(
-        state_client,
-        config.limits.max_commitments_per_slot,
-        config.limits.max_committed_gas_per_slot,
-    )
-    .await?;
+    let mut execution_state = ExecutionState::new(state_client, config.limits.clone()).await?;
 
     let mevboost_client = MevBoostClient::new(config.mevboost_url.clone());
     let beacon_client = BeaconClient::new(config.beacon_api_url.clone());
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -28,7 +28,7 @@ async fn main() -> eyre::Result<()> {
     let signer = Signer::new(config.private_key.clone().unwrap());
 
     let state_client = StateClient::new(config.execution_api_url.clone());
-    let mut execution_state = ExecutionState::new(state_client, config.limits.clone()).await?;
+    let mut execution_state = ExecutionState::new(state_client, config.limits).await?;
 
     let mevboost_client = MevBoostClient::new(config.mevboost_url.clone());
     let beacon_client = BeaconClient::new(config.beacon_api_url.clone());
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -28,8 +28,12 @@ async fn main() -> eyre::Result<()> {
     let signer = Signer::new(config.private_key.clone().unwrap());
 
     let state_client = StateClient::new(config.execution_api_url.clone());
-    let mut execution_state =
-        ExecutionState::new(state_client, config.limits.max_commitments_per_slot).await?;
+    let mut execution_state = ExecutionState::new(
+        state_client,
+        config.limits.max_commitments_per_slot,
+        config.limits.max_committed_gas_per_slot,
+    )
+    .await?;
 
     let mevboost_client = MevBoostClient::new(config.mevboost_url.clone());
     let beacon_client = BeaconClient::new(config.beacon_api_url.clone());
--- a/bolt-sidecar/src/builder/template.rs
+++ b/bolt-sidecar/src/builder/template.rs
@@ -111,6 +111,18 @@ impl BlockTemplate {
             .fold(0, |acc, sc| acc + sc.message.constraints.len())
     }
 
+    /// Returns the committed gas in the block template.
+    #[inline]
+    pub fn committed_gas(&self) -> U256 {
+        self.signed_constraints_list
+            .iter()
+            .fold(U256::ZERO, |acc, sc| {
+                acc + sc.message.constraints.iter().fold(U256::ZERO, |acc, c| {
+                    acc + max_transaction_cost(&c.transaction)
+                })
+            })
+    }
+
     /// Returns the blob count of the block template.
     #[inline]
     pub fn blob_count(&self) -> usize {
--- a/bolt-sidecar/src/builder/template.rs
+++ b/bolt-sidecar/src/builder/template.rs
@@ -113,14 +113,14 @@ impl BlockTemplate {
 
     /// Returns the committed gas in the block template.
     #[inline]
-    pub fn committed_gas(&self) -> U256 {
-        self.signed_constraints_list
-            .iter()
-            .fold(U256::ZERO, |acc, sc| {
-                acc + sc.message.constraints.iter().fold(U256::ZERO, |acc, c| {
-                    acc + max_transaction_cost(&c.transaction)
-                })
-            })
+    pub fn committed_gas(&self) -> u64 {
+        self.signed_constraints_list.iter().fold(0, |acc, sc| {
+            acc + sc
+                .message
+                .constraints
+                .iter()
+                .fold(0, |acc, c| acc + &c.transaction.gas_limit())
+        })
     }
 
     /// Returns the blob count of the block template.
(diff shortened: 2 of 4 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -145,6 +150,7 @@ impl<C: StateFetcher> ExecutionState<C> {
     pub async fn new(
         client: C,
         max_commitments_per_slot: NonZero<usize>,
+        max_committed_gas_per_slot: NonZero<u64>,
     ) -> Result<Self, TransportError> {
```
merklefruit: can we pass `Limits` as argument here instead of both the values contained in Limits?

thedevbirb:
Nice work! Can you please add some tests?

bolt-sidecar/src/builder/template.rs
```diff
@@ -111,6 +111,18 @@ impl BlockTemplate {
...
+    #[inline]
+    pub fn committed_gas(&self) -> U256 {
+        self.signed_constraints_list
+            .iter()
+            .fold(U256::ZERO, |acc, sc| {
+                acc + sc.message.constraints.iter().fold(U256::ZERO, |acc, c| {
+                    acc + max_transaction_cost(&c.transaction)
+                })
+            })
+    }
```
merklefruit: committed gas should keep track of the amount of gas units used by all commitments, therefore it should sum the `gas_limit` of all transactions rather than the `max_transaction_cost`

bolt-sidecar/src/state/execution.rs
```diff
@@ -200,12 +207,23 @@ impl<C: StateFetcher> ExecutionState<C> {
...
+            }
+
+            // Check if the committed gas exceeds the maximum
+            if template.committed_gas().to::<u64>() >= max_committed_gas_per_slot {
+                return Err(ValidationError::MaxCommittedGasReachedForSlot(
+                    self.slot,
+                    max_committed_gas_per_slot,
                 ));
             }
         }
```
thedevbirb: Here I think you should check if 
`template.committed_gas() + tx.gas_limit >= max_committed_gas_per_slot`
otherwise with the previous request I can ask for huge gas limit and overcome the max value.
Does it make sense?

bolt-sidecar/src/state/execution.rs
```diff
@@ -801,9 +812,11 @@ mod tests {
...
         let client = StateClient::new(anvil.endpoint_url());
 
-        let max_comms = NonZero::new(10).unwrap();
-        let max_gas = NonZero::new(10_000_000).unwrap();
-        let mut state = ExecutionState::new(client.clone(), max_comms, max_gas).await?;
+        let limits: Limits = Limits {
+            max_commitments_per_slot: NonZero::new(10).unwrap(),
+            max_committed_gas_per_slot: NonZero::new(10_000_000).unwrap(),
+        };
+        let mut state = ExecutionState::new(client.clone(), limits).await?;
```
merklefruit: you can also do `ExecutionState::new(client.clone(), Limits::default())` if you want - it's ok like this tho!

merklefruit (approved):
Great! LGTM so far

bolt-sidecar/src/state/execution.rs
namn-grg: Yes good point! Then I think we should also do the `tx.gas_limit >= max_committed_gas_per_slot` check in case no template is there right? updated

bolt-sidecar/src/state/execution.rs
namn-grg: good point!

bolt-sidecar/src/state/execution.rs
```diff
@@ -210,6 +214,24 @@ impl<C: StateFetcher> ExecutionState<C> {
...
+                ));
+            }
+        } else if req.tx.gas_limit() >= self.max_committed_gas_per_slot.get() {
+            // Check if the committed gas limit itself exceeds the maximum when template is None
+            return Err(ValidationError::MaxCommittedGasReachedForSlot(
+                self.slot,
+                self.max_committed_gas_per_slot.get(),
+            ));
+        }
+
```
namn-grg: This check is for the case when no template is built and the request itself exceeds the limit

bolt-sidecar/bin/sidecar.rs
```diff
@@ -28,8 +28,7 @@ async fn main() -> eyre::Result<()> {
     let signer = Signer::new(config.private_key.clone().unwrap());
 
     let state_client = StateClient::new(config.execution_api_url.clone());
-    let mut execution_state =
-        ExecutionState::new(state_client, config.limits.max_commitments_per_slot).await?;
+    let mut execution_state = ExecutionState::new(state_client, config.limits.clone()).await?;
```
merklefruit: nit: if we derive `Copy` on `Limits`, we can get rid of this clone. This doesn't change much but it's a tiny bit cleaner!

bolt-sidecar/src/state/execution.rs
```diff
@@ -113,6 +117,8 @@ pub struct ExecutionState<C> {
     chain_id: u64,
     /// The maximum number of commitments per slot.
     max_commitments_per_slot: NonZero<usize>,
+    /// The maximum committed gas per slot.
+    max_committed_gas_per_slot: NonZero<u64>,
     /// The KZG settings for validating blobs.
```
namn-grg: Is it better to make them into a single field - `limits`?

bolt-sidecar/src/state/execution.rs
thedevbirb: This can be refactored a bit! Let me know wdyt
```suggestion
        // Check if the committed gas exceeds the maximum
        let template_committed_gas = self.get_block_template(target_slot).map(|t| t.committed_gas()).unwrap_or(0);
        if template_committed_gas + req.tx.gas_limit() >= self.max_committed_gas_per_slot.get() {
            return Err(ValidationError::MaxCommittedGasReachedForSlot(
                self.slot,
                self.max_committed_gas_per_slot.get(),
            ));
        }
```

bolt-sidecar/src/state/execution.rs
merklefruit: yes in my opinion

bolt-sidecar/src/state/execution.rs
```diff
@@ -210,6 +214,24 @@ impl<C: StateFetcher> ExecutionState<C> {
             }
         }
 
+        // Check if the committed gas exceeds the maximum
+        if let Some(template) = self.get_block_template(target_slot) {
+            // Check if the committed gas exceeds the maximum
```
merklefruit:
```suggestion
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -210,6 +214,24 @@ impl<C: StateFetcher> ExecutionState<C> {
...
+            if template.committed_gas() + req.tx.gas_limit()
+                >= self.max_committed_gas_per_slot.get()
+            {
+                return Err(ValidationError::MaxCommittedGasReachedForSlot(
+                    self.slot,
+                    self.max_committed_gas_per_slot.get(),
+                ));
+        // Check if the committed gas exceeds the maximum
```
merklefruit:
```suggestion
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -210,6 +214,18 @@ impl<C: StateFetcher> ExecutionState<C> {
             }
         }
 
+        // Check if the committed gas exceeds the maximum
+        let template_committed_gas = self
+            .get_block_template(target_slot)
+            .map(|t| t.committed_gas())
+            .unwrap_or(0);
```
merklefruit, line 218: awesome!

merklefruit (approved):

bolt-sidecar/src/state/execution.rs
namn-grg: done

thedevbirb (approved):
Let's gooo

PR #151, opened 2024-07-19
feat: small sidecar changes

merklefruit:
- some minor code quality improvements in bolt-sidecar


Justfile
thedevbirb, line 1: The `setup_remote_builder.sh` file setups the remote builder context, however the build commands have not been changed to use such context if available.

merklefruit: yeah, I'm not sure if I want to make it the default, also because then it would only work locally for us inside the Tailnet. I am leaning towards leaving this here and manually adding the --context flag when needed

also we don't always need this...

wdyt?

thedevbirb: We could make the build command take a parameter which defaults to the local context but uses another if specified. Btw building times have been perfectly fine for me once caching started working again.

merklefruit: Ok!

merklefruit:
Decided to revert the docker context script, bringing only small sidecar changes to this PR.

thedevbirb (approved):

PR #158, opened 2024-07-23
feat(registry): add getallregistrants

### LINKED ISSUE #157: feat(contracts): add a way to query all registered validators in Bolt-registry
Currently, there is no good way to query all registered operators. Something like an operators[] array would be just fine, as each operator already has its list of active validator indexes

namn-grg:
fixes #157 

## Summary by CodeRabbit

- **New Features**
	- Introduced a method to retrieve all active registrants associated with operators.
	- Added a public array to track active operator addresses for improved operator management.
- **Improvements**
	- Enhanced the accessibility of the operator status check, allowing internal and external calls.

### DIFF
--- a/bolt-contracts/src/contracts/BoltRegistry.sol
+++ b/bolt-contracts/src/contracts/BoltRegistry.sol
@@ -50,6 +50,8 @@ contract BoltRegistry is IBoltRegistry {
             metadata
         );
 
+        operators.push(msg.sender);
+
         // Set the delegations
         for (uint256 i = 0; i < validatorIndexes.length; i++) {
             delegations[validatorIndexes[i]] = msg.sender;
--- a/bolt-contracts/src/contracts/BoltRegistry.sol
+++ b/bolt-contracts/src/contracts/BoltRegistry.sol
@@ -97,6 +97,15 @@ contract BoltRegistry is IBoltRegistry {
             revert CooldownNotElapsed();
         }
 
+        // Remove operator from the operators array
+        for (uint256 i = 0; i < operators.length; i++) {
+            if (operators[i] == msg.sender) {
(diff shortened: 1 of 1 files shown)


bolt-contracts/src/contracts/BoltRegistry.sol
```diff
@@ -92,6 +97,15 @@ contract BoltRegistry is IBoltRegistry {
             revert CooldownNotElapsed();
         }
 
+        // Remove operator from the operators array
+        for (uint256 i = 0; i < operators.length; i++) {
+            if (operators[i] == msg.sender) {
+                operators[i] = operators[operators.length - 1];
```
mempirate, line 103: Is this how you remove in solidity? Can't you just `delete operators[i]`?

merklefruit: delete will just set that value to 0 instead of resizing the array

e.g. before and after:  [1, 2, 3, 4, 5] -> [1, 2, 0, 4, 5]
where we want: [1, 2, 4, 5]

so this is correct

thedevbirb: With this example, assume `i` equals `2`. Then `operators[i]` will be set to `operators[4]` which is 5, so we have the list `[1,2,5,4,5]`, then we pop so we're left with `[1,2,5,4]`. Am I missing something?

merklefruit: No you are totally right, I should have said "where we want [1, 2, 5, 4]" my bad.
We don't care about the order of these items

namn-grg: @merklefruit has explained this right! This should be correct imo

PR #161, opened 2024-07-24
Implement new commitments-API

### LINKED ISSUE #156: Commitments-API spec implementation
# Context
Implement the commitments-API spec as defined below in the Bolt sidecar.

# Specifications
### Description
Requests an inclusion preconfirmation for one or more transactions. If `len(txs) > 1`, the array will be considered an atomic bundle.

> [!IMPORTANT]  
> **Creating the signature**: the signature should come from signing the message digest. The message digest
> is the following: 
> ```rust
> let message_digest = {
>      let mut data = Vec::new();
>      // First field is the concatenation of all the transaction hashes
>      data.extend_from_slice(&transactions.map(|tx| tx.hash().as_slice()).concat());
>      // Second field is the little endian encoding of the target slot
>      data.extend_from_slice(&target_slot.to_le_bytes());
>      keccak256(data)
> };
> ```
> The signature should be stored in the `X-Bolt-Signature` header as a hex-encoded string. It will be used for authentication purposes.

### Parameters
| Method | Content-Type | X-Bolt-Signature |
| -------- | ------------- | ----------------- |
| POST | `application/json` | <ecdsa_signature> |

### Body
```json5
{
  "jsonrpc": "2.0",
  "id": "1",
  "method": "bolt_requestInclusion",
  "params": [
    {
      txs,         // Array[String], A list of hex-encoded "raw" transactions. If type 3, must include blobs.
      targetSlot  // String, a hex-encoded string of the target slot. 
    }
  ]
}
```

mempirate:
Closes #156 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `commitments` module, enhancing API capabilities for commitment requests.
	- Added JSON-RPC support for handling commitment requests, improving API communication.
	- Implemented new structures for handling transaction commitments and signatures, enhancing the commitment system.

- **Bug Fixes**
	- Improved error handling and validation logic for commitment requests, ensuring more reliable operations.

- **Style**
	- Reformatted dependency entries for better readability in configuration files.

- **Chores**
	- Updated module structures and public API exports for improved maintainability.


bolt-sidecar/src/state/execution.rs
```diff
@@ -269,89 +269,95 @@ impl<C: StateFetcher> ExecutionState<C> {
...
-
-                (
-                    nonce_diff_acc + nonce_diff,
-                    balance_diff_acc.saturating_add(balance_diff),
-                    u64::max(highest_slot, slot),
-                )
-            },
-        );
-
-        if target_slot < highest_slot {
```
mempirate, line 295: @thedevbirb unsure why this error is only thrown when the target slot is smaller than the highest slot in the template. It should be thrown every time the target slot is lower than the current one, which is stored in `self.slot` right?

thedevbirb: In general yes we should also check that `target_slot >= self.slot`. I thought we were doing that already (I could be wrong)

The idea of that check is the following: assume I have nonce $N$, I sent a preconf for slot $N+2$ and then a preconf for slot $N+1$. In theory, the latter is valid but it invalidates the preconf for slot $N+2$. To avoid managing these edge cases, we simply require that preconfirmations in a increasing order of `target_slot`.
Does it make sense in your opinion? cc @merklefruit since we discussed about this together a while ago

mempirate:
@coderabbitai review

bolt-sidecar/bin/sidecar.rs
```diff
@@ -27,14 +30,22 @@ async fn main() -> eyre::Result<()> {
...
     let state_client = StateClient::new(config.execution_api_url.clone());
     let mut execution_state = ExecutionState::new(state_client, config.limits).await?;
 
     let mevboost_client = MevBoostClient::new(config.mevboost_url.clone());
     let beacon_client = BeaconClient::new(config.beacon_api_url.clone());
 
     let (api_events, mut api_events_rx) = mpsc::channel(1024);
-    let shutdown_tx = start_rpc_server(&config, api_events).await?;
+
+    let mut api_server = CommitmentsApiServer::new(format!("0.0.0.0:{}", config.rpc_port));
```
merklefruit, line 44: nit: can we just pass the rpc_port as argument?

bolt-sidecar/bin/sidecar.rs
```diff
@@ -65,44 +76,52 @@ async fn main() -> eyre::Result<()> {
...
-                let res = serde_json::to_value(signed_constraints).map_err(Into::into);
-                let _ = response_tx.send(res).ok();
+                // Create a commitment by signing the request with the commitment signer
+                match request.commit_and_sign(&commitment_signer).await {
+                    Ok(commitment) => {
+                        let _ = response.send(Ok(commitment));
+                    },
+                    Err(e) => {
+                        tracing::error!(err = ?e, "Failed to sign commitment");
+                        let _ = response.send(Err(spec::Error::Internal));
```
merklefruit, line 121: can we provide a more detailed error for users here?

bolt-sidecar/bin/sidecar.rs
```diff
@@ -163,7 +182,7 @@ async fn main() -> eyre::Result<()> {
             },
             Ok(_) = tokio::signal::ctrl_c() => {
                 tracing::info!("Received SIGINT, shutting down...");
-                shutdown_tx.send(()).await.ok();
+                // shutdown_tx.send(()).await.ok();
```
merklefruit:
```suggestion
```

bolt-sidecar/bin/sidecar.rs
```diff
@@ -27,14 +30,22 @@ async fn main() -> eyre::Result<()> {
...
     let beacon_client = BeaconClient::new(config.beacon_api_url.clone());
 
     let (api_events, mut api_events_rx) = mpsc::channel(1024);
-    let shutdown_tx = start_rpc_server(&config, api_events).await?;
+
+    let mut api_server = CommitmentsApiServer::new(format!("0.0.0.0:{}", config.rpc_port));
+
+    api_server.run(api_events).await?;
+
+    // let shutdown_tx = start_rpc_server(&config, api_events).await?;
```
merklefruit:
```suggestion
```

bolt-sidecar/src/api/commitments/server.rs
```diff
@@ -0,0 +1,373 @@
...
+            })),
+        }
+    }
+
+    /// Creates the server with the given address and shutdown signal.
+    pub fn with_shutdown<A: ToSocketAddrs>(
+        self,
+        addr: A,
+        signal: impl Future<Output = ()> + Send + 'static,
+    ) -> Self {
```
merklefruit:
```suggestion
    pub fn with_shutdown<A: ToSocketAddrs>(self, addr: A, signal: S) -> Self
    where 
        A: ToSocketAddrs,
        S: Future<Output = ()> + Send + 'static
    {
```

bolt-sidecar/src/api/commitments/server.rs
```diff
@@ -0,0 +1,373 @@
...
+                let response = JsonResponse {
+                    id: payload.id,
+                    result: serde_json::to_value(inclusion_commitment).unwrap(),
+                    ..Default::default()
+                };
+
+                Ok(Json(response))
+            }
+            _ => {
+                tracing::error!("Unknown method: {}", payload.method);
```
merklefruit:
```suggestion
            other => {
                tracing::error!("Unknown method: {}", other);
```

bolt-sidecar/src/api/commitments/spec.rs
```diff
@@ -0,0 +1,125 @@
...
+    #[error("{0}")]
+    ValidationFailed(String),
+    /// Duplicate request.
+    #[error("Duplicate request")]
+    Duplicate,
+    /// Internal server error.
+    #[error("Internal server error")]
+    Internal,
+    /// Missing signature.
+    #[error("Missing X-Bolt-Signature header")]
```
merklefruit:
```suggestion
    #[error("Missing '{SIGNATURE_HEADER}' header")]
```

merklefruit (approved):
LGTM, a few nits

bolt-sidecar/bin/sidecar.rs
mempirate: In the future we should allow proposers to specify hosts as well, i.e. `localhost` or `0.0.0.0` so they don't expose ports if they don't need to.

bolt-sidecar/bin/sidecar.rs
mempirate: This will only fail in case the proposer tries to sign with an invalid signature, should not be exposed to the user

namn-grg (approved):
Great work LGTM!

thedevbirb (approved):
Awesome!

bolt-sidecar/bin/sidecar.rs
merklefruit: ah good point, ok

bolt-sidecar/bin/sidecar.rs
merklefruit: ok gotcha

PR #162, opened 2024-07-24
fix(devnet): better support of multiple preconfs in blocks

### LINKED ISSUE #148: bug: merkle multiproof verification
As outlined in #146, merkle multiproofs are currently failing with 3+ constraints per slot.
We are currently investigating. this issue will keep track of the results and proposed solutions.

This issue affects `mev-boost-relay`, `builder` and `mev-boost`.

thedevbirb:
This PR introduces a better support for multiple (>2) preconfirmations in a block, which was mostly not working before. This has been achieved with the following steps:
1. The dependency https://github.com/ferranbt/fastssz has been updated after the fix to multiproof verifcation code: https://github.com/ferranbt/fastssz/pull/173
2. Allow devnet spammer to send multiple preconfs at the same time using the `--count` flag; an appropriate `justfile` command has been introduced
3. A deterministic ordering logic for unindexed preconfirmations has been introduced: such transactions are put at the bottom of the block but in order to satisfy protocol rules they have been ordered by nonce ascending, and in case two preconfs have the same nonce, their hashes are compared.
4. This ordering logic influences the way merkle multiproofs are created and verified. In particular, it is enforced that generalised indices must be sorted in ascending order, and both the relay and MEV-Boost fork order the constraints according to rule above when verifying

The code has been tested in the follwing scenarios:
- regular transactions, remote builder
- 6 blob transactions with 1 blob each, remote builder
- regular transactions, local builder
- 6 blob transactions with 1 blob each, local builder

closes #148 

## Summary by CodeRabbit

- **New Features**
  - Enhanced preconfirmation commands to allow specification of the number of preconfirmations to send.
  - Introduced support for sending multiple transactions simultaneously via a new command-line option.

- **Enhancements**
  - Increased default capacity for commitments per slot to improve performance.
  - Updated gas limits for transactions to reflect more realistic consumption.

- **Dependency Updates**
  - Upgraded Go version used in Dockerfiles and modules.
  - Updated and added dependencies in various modules for improved functionality.

- **Refactor**
  - Improved constraints processing with new structured types, enhancing clarity and performance in the codebase.

- **Tests**
  - Introduced a new test for validating the parsing of transaction constraints to ensure functionality integrity.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -102,16 +102,7 @@ dora:
   fi
 
 # manually send a preconfirmation to the bolt devnet
-send-preconf:
-	cd bolt-spammer && RUST_LOG=info cargo run -- \
-		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
-		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
-		--bolt-sidecar-url http://$(kurtosis port print bolt-devnet mev-sidecar-api api)  \
-		--private-key 53321db7c1e331d93a11a41d16f004d7ff63972ec8ec7c25db329728ceeb1710 \
-		--slot head
-
-# manually send a preconfirmation to the bolt devnet
-send-preconf-count count:
+send-preconf count='1':
 	cd bolt-spammer && RUST_LOG=info cargo run -- \
 		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
 		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
@@ -121,17 +112,7 @@ send-preconf-count count:
 		--count {{count}}
 
 # manually send a blob preconfirmation to the bolt devnet
-send-blob-preconf:
-	cd bolt-spammer && RUST_LOG=info cargo run -- \
-		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
-		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
-		--bolt-sidecar-url http://$(kurtosis port print bolt-devnet mev-sidecar-api api)  \
-		--private-key 53321db7c1e331d93a11a41d16f004d7ff63972ec8ec7c25db329728ceeb1710 \
-		--slot head \
-		--blob
-
-# manually send a blob preconfirmation to the bolt devnet
-send-blob-preconf-count count:
+send-blob-preconf count='1':
 	cd bolt-spammer && RUST_LOG=info cargo run -- \
 		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
 		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
--- a/Justfile
+++ b/Justfile
@@ -110,6 +110,16 @@ send-preconf:
 		--private-key 53321db7c1e331d93a11a41d16f004d7ff63972ec8ec7c25db329728ceeb1710 \
 		--slot head
 
+# manually send a preconfirmation to the bolt devnet
+send-preconf-count count:
+	cd bolt-spammer && RUST_LOG=info cargo run -- \
+		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
+		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
+		--bolt-sidecar-url http://$(kurtosis port print bolt-devnet mev-sidecar-api api)  \
+		--private-key 53321db7c1e331d93a11a41d16f004d7ff63972ec8ec7c25db329728ceeb1710 \
+		--slot head \
+		--count {{count}}
+
 # manually send a blob preconfirmation to the bolt devnet
 send-blob-preconf:
 	cd bolt-spammer && RUST_LOG=info cargo run -- \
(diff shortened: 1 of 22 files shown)


Justfile
```diff
@@ -110,6 +110,16 @@ send-preconf:
...
 
+# manually send a preconfirmation to the bolt devnet
+send-preconf-count count:
+	cd bolt-spammer && RUST_LOG=info cargo run -- \
+		--provider-url $(kurtosis port print bolt-devnet el-1-geth-lighthouse rpc) \
+		--beacon-client-url $(kurtosis port print bolt-devnet cl-1-lighthouse-geth http) \
+		--bolt-sidecar-url http://$(kurtosis port print bolt-devnet mev-sidecar-api api)  \
+		--private-key 53321db7c1e331d93a11a41d16f004d7ff63972ec8ec7c25db329728ceeb1710 \
+		--slot head \
+		--count {{count}}
```
merklefruit: can we have just one recipe with `count=1` as default? (same with blob command below)

bolt-spammer/src/main.rs
```diff
@@ -53,45 +55,47 @@ async fn main() -> Result<()> {
...
-    let client = reqwest::Client::new();
-    let response = client
-        .post(&opts.bolt_sidecar_url)
-        .header("content-type", "application/json")
-        .body(serde_json::to_string(&request)?)
-        .send()
-        .await?;
-
-    info!("Response: {:?}", response.text().await?);
+    for i in 0..opts.count {
```
merklefruit, line 58: there is no delay between spamming? should we do something like 100 ms? just an idea

mev-boost/server/constraints.go
```diff
@@ -95,16 +98,64 @@ func (c *ConstraintCache) AddInclusionConstraints(slot uint64, constraints []*Co
...
+		Index *uint64
+		Tx    *types.Transaction
+	}
+)
+
+// ParseConstraintsDecoded receives a map of constraints and
+// - creates a slice of constraints sorted by index
+// - creates a slice of constraints without index sorted by nonce and hash
+// Returns the concatenation of the slices
+func ParseConstraintsDecoded(constraints HashToConstraintDecoded) []*ConstraintDecoded {
```
merklefruit, line 127: can you add a test for this fn? either in the relay or mev-boost is fine

merklefruit (approved):
Great! A couple small changes, and then we can test this on Helder IMO!

bolt-spammer/src/main.rs
thedevbirb: For the devnet I think there is no reason to add a delay, but yeah I should bring these changes for helder testnet and it might be worth to add configurable delay between preconfs, with a 100ms default

thedevbirb: Actually, should I do that for Helder or wait for #163 to be done?

Justfile
thedevbirb: Good call, for some reasons I thought it was not possible with Justfiles

bolt-spammer/src/main.rs
merklefruit: waiting is fine, it's not a priority

mev-boost/server/constraints.go
thedevbirb: Done! Lmk if it's good enough

merklefruit: beautiful!

merklefruit (approved):
looks gucci to me

builder/builder/utils.go
```diff
@@ -156,8 +156,11 @@ func EmitBoltDemoEvent(message string) {
 
 func CalculateMerkleMultiProofs(
 	payloadTransactions types.Transactions,
-	constraints types.HashToConstraintDecoded,
+	HashToConstraintDecoded types.HashToConstraintDecoded,
```
mempirate, line 159: nit but why not `constraints` instead of same name as the type? In general I think shorter variables improve readability

thedevbirb: Just because `constraints` is the name I've given to the constraints array shortly after, so I've changed this to something that resembles more closely its type. However this is not set in stone! Any suggestion is welcome

mev-boost-relay/services/api/proofs.go
```diff
@@ -20,37 +19,23 @@ var (
...
-		// parsed without the blob here to avoid unmarshalling and marshalling again
-		transaction := new(types.Transaction)
-		err := transaction.UnmarshalBinary(constraint.Tx)
-		if err != nil {
-			log.WithError(err).Error("error unmarshalling transaction while verifying proofs")
-			return err
-		}
-
-		withoutBlob, err := transaction.WithoutBlobTxSidecar().MarshalBinary()
+		withoutBlob, err := constraint.Tx.MarshalBinary()
```
mempirate: If this is without blob, how do you get the encoding with blob?

mempirate (approved):
LGTM!

mev-boost-relay/services/api/proofs.go
thedevbirb: Good catch, actually when calling this functions the constraints txs are already forwarded without blobs (see https://github.com/chainbound/bolt/pull/162/files/880cc52196201062f1021ee9db549b93109ed00a#diff-f03b62c88edc3b36b56c7ee1b08c7f6670b22e21c52d18430f9f1d7f3eff7534), however the variable name has been left the same which is a bit misleading

mev-boost-relay/services/api/service.go
```diff
@@ -2158,18 +2158,17 @@ func (api *RelayAPI) updateRedisBidWithProofs(
...
-				// TODO: just compute the hash instead of decoding the entire tx just for this.
 				decoded := new(types.Transaction)
 				if err := decoded.UnmarshalBinary(constraint.Tx); err != nil {
 					api.log.WithError(err).Error("could not decode transaction")
 					api.RespondError(opts.w, http.StatusBadRequest, "could not decode transaction")
 					return nil, nil, false
 				}
 				api.log.Infof("Decoded tx hash %s", decoded.Hash().String())
-				constraints[phase0.Hash32(decoded.Hash().Bytes())] = constraint
+				constraints[decoded.Hash()] = &ConstraintDecoded{Tx: decoded.WithoutBlobTxSidecar(), Index: constraint.Index}
```
thedevbirb, line 2171: Constraints for the proofs are already saved without blobs here

PR #166, opened 2024-07-26
fix(mevboost, relay): ordering logic for multiproofs

thedevbirb:
After some internal discussion we figured out that our MEV-Boost and relay fork don't need to sort the unindexed constraints like builder does, see #162 for context, but it is sufficient that the [`InclusionProof`](https://chainbound.github.io/bolt-docs/api/builder#ethv1builderheader_with_proofsslotparent_hashpubkey) struct satisfies the following property:
> _The `TransactionHashes` field must be following the same order of `GeneralizedIndexes`_

An example is the following: if on a block $B$ I have two preconfirmed transactions $T_1$ and $T_2$ at index $i_1$, $i_2$ respectively and $g_1$, $g_2$ are their generalized indexes, then the proof data must be:
```text
TransactionHashes: [Hash(T_1), Hash(T_2)],
GeneralizedIndexes: [g_1, g_2]
```
or, equivalently
```text
TransactionHashes: [Hash(T_2), Hash(T_1)],
GeneralizedIndexes: [g_2, g_1]
```


mev-boost-relay/services/api/proofs.go
```diff
@@ -50,17 +61,14 @@ func verifyInclusionProof(log *logrus.Entry, transactionsRoot phase0.Root, proof
 		}
 
 		leaves[i] = txHashTreeRoot[:]
+		indexes[i] = int(proof.GeneralizedIndexes[i])
```
mempirate, line 64: Should we be careful with converting from uin64 to int64 here?

mempirate (approved):
LGTM, let's test this and then merge

mev-boost-relay/services/api/proofs.go
merklefruit: This is safe because max i64 is 9,223,372,036,854,775,807 and all generalized indexes for transactions are capped at 2^22 -1 = 4,194,303 by consensus specs

thedevbirb: For the multiproof verification it is required that `indexes` is a `[]int`. Given that generalized indexes will reasonably be small numbers (in the order of $2^{21}$ in our use case), we should not worry too much about it

merklefruit (approved):

PR #167, opened 2024-07-26
feat(devnet): blob express lane PoC

### LINKED ISSUE #164: Blob Express Lane integration
> _Blob Express Lane (BXL) is a product built on Bolt that aims to fix some of the inefficiencies in the blob market. This would result in more predictable inclusion & costs for rollups, less hidden latency costs for builders, and more revenue for proposers._

In essence, there is an optimisation in the PBS pipeline that can be done with regards to blob preconfirmations: since both MEV-Boost and relays already know a preconfirmed blob sidecar, there is no specific reason for a builder to forward them again the bid, by saving up some latency costs.
This has additional benefits that are explored in an upcoming article.

From an implementation perspective, these are the steps required for the integration of the BXL:
1. In MEV-Boost, upon receiving a blob constraint, the blob sidecar is integrated in a local [`BlobsBundle`](https://github.com/ethereum/consensus-specs/blob/2218c94e25411c3b25a129453e10962f01226991/specs/deneb/validator.md#blobsbundle), saved in a cache. The constraint is then forwarded to the relay as usual with the full blob contents
2. In the relay, upon receiving a blob constraint, the same thing is done. It will also save the `blob_versioned_hash` associated with every KZG commitment for later use
3. The builder will receive the blob constraint. The full content is needed because of execution validation purposes. However, the builder will just strip out the blob sidecar from constrained type 3 transactions before sending the bid. Given all constraints are forwarded with a deadline which is ~8 into the slot, the latency added is negligible.
4. When the relay receives a bid, it should check whether all type 3 transactions in it have an associated sidecar in the cache. This check is done by comparing the `blob_versioned_hashes` field of the transactions with the ones saved in the cache. If the check passes, the bid is saved as usual.
5. When the `getPayload` call comes the relay will publish the assembled block to the network but it can also skip forwarding the preconfirmed blobs to MEV-Boost, since they are already available
6. In MEV-Boost, upon receiving the `getPayload` response from the relay the payload is assembled with the local blobs bundle, and the block is sent to the beacon node.

thedevbirb:
Integrates the Blob Express Lane as described in #164

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -56,11 +56,6 @@ sidecar-logs:
     @id=$(docker ps -n 100 | grep sidecar | awk -F' ' '{print $1}') && \
     docker logs -f $id
 
-# show the logs for the bolt devnet beacon node
-beacon-logs:
-    @id=$(docker ps -n 100 | grep 'cl-1-lighthouse-geth' | awk -F' ' '{print $1}') && \
-    docker logs -f $id
-
 # show the logs for the bolt devnet for beacon node
 beacon-logs:
     @id=$(docker ps -n 100 | grep 'cl-1-lighthouse-geth' | awk -F' ' '{print $1}') && \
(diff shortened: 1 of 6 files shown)


thedevbirb:
Closing this to keep the repo clean, it will be there to serve as a reference if needed.

PR #169, opened 2024-07-29
fix(sidecar): process pending bundle diffs

merklefruit:
### Overview

When validating requests containing multiple transactions, we need to take into account the previous transactions in the bundle as increasing the nonce (and decreasing the balance) of the account that sent it, even if they haven't been committed yet to the template. 

This fixes the issue by introducing two maps to keep track of the pending state due to these requests. 

Note that if anything in the request fails validation, the whole bundle won't be committed, as expected.

### Tests
- currently testing from this branch in release candidate `v0.1.3-alpha.rc2`
- multiple type 2 txs: Included: https://dora.helder-devnets.xyz/slot/273556

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -11,7 +11,7 @@ use thiserror::Error;
 
 use crate::{
     builder::BlockTemplate,
-    common::{calculate_max_basefee, validate_transaction},
+    common::{calculate_max_basefee, max_transaction_cost, validate_transaction},
     config::Limits,
     primitives::{AccountState, CommitmentRequest, SignedConstraints, Slot},
 };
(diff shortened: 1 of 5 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -326,12 +337,30 @@ impl<C: StateFetcher> ExecutionState<C> {
...
+                    .saturating_add(nonce_diff)
+                    .saturating_add(*sender_nonce_diff),
+
+                balance: account_state
+                    .balance
+                    .saturating_sub(balance_diff)
+                    .saturating_sub(*sender_balance_diff),
+
+                // TODO(nico): what if this changes in the middle of the request?
                 has_code: account_state.has_code,
```
thedevbirb: I don't think this can happen even with EIP-7702. What are your doubts about it?

thedevbirb (approved):
Thanks for the PR! The changes make sense. If you have time, can please add some unit tests with bundles?

bolt-sidecar/src/state/execution.rs
merklefruit: Yeah, after thinking more about it, I am convinced it can't happen

merklefruit: removed

mempirate (approved):
Lgtm!

PR #172, opened 2024-07-29
feat(sidecar): add `bolt_getVersion` RPC method

### LINKED ISSUE #171: chore(sidecar): expose a version endpoint
As part of our next releases, we are going to start using the Sidecar's `CARGO_PKG_VERSION` to reliably keep track of versions of the software that node operators are running. Ideally, this should also be available as a header on a per-request basis. 

This issue consists in adding this functionality with both a `version` API endpoint and an HTTP header `x-sidecar-version`.

A more generic approach could be to introduce a more elaborate `Capabilities` enum along with the version, similar to Ethereum's engine api specs, but I think this is overkill for our current needs.

merklefruit:
Added an endpoint to fetch the version of the Bolt-sidecar, 
so that external consumers know which functionalities are implemented. 

Also added a couple debug logs to inspect some issues with bundle inclusion preconfirmations on Helder.

- closes #171


mempirate (approved):
LGTM!

PR #179, opened 2024-08-01
fix(sidecar): bundle verification logic

### LINKED ISSUE #174: fix(sidecar): investigate `NonceTooLow` error when sending requests with RPC through Metamask
It's not clear if Metamask is spamming transactions to the RPC or if there is a validation bug when trying to send multiple inclusion requests to the same sidecar. For now this issue is a reminder to investigate this behavior

thedevbirb:
In the sidecar we've recently introduced bundle support, which need special nonce and balance checks when verified. 

In particular, when processing the next transaction in the bundle we should increment the user nonce and decrease its balance, however this operationis done before the transaction validation and not at the end of the loop, causing an off-by-one error in nonce checks (`NonceTooLow`)

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -358,10 +358,6 @@ impl<C: StateFetcher> ExecutionState<C> {
                 has_code: account_state.has_code,
             };
 
-            // Increase the bundle nonce and balance diffs for this sender for the next iteration
-            *sender_nonce_diff += 1;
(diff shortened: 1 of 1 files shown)


merklefruit:
Great catch! (Pic unrelated)

Off_By_One

merklefruit (approved):

thedevbirb:
Proof of ItWorks™️ https://dora.helder-devnets.xyz/slot/296234: three preconfs from the same address

PR #181, opened 2024-08-02
feat(sidecar): added driver on higher register (`tokio::select`)

merklefruit:
This PR introduces a refactor of the `sidecar` binary to streamline all components that go in the initialization and event loop phase of operating the sidecar.

In particular, this implementation differs from what attempted in #133 in not using a manual `Future` but instead still relying on `tokio::select!` to progress the loop. This simplifies things a lot, because of the following async tasks:

1. updating the consensus and execution head is an async operation (it requires fetching from both EL and CL)
2. validating the execution state of a request is an async operation (it requires fetching state from the EL)

This could be overcome in the future by transforming the `ExecutionState` and `ConsensusState` containers into actors with their own loops to respond to requests in channels instead of awaiting.

### DIFF
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -1,188 +1,22 @@
-use std::time::Duration;
-
-use alloy::{rpc::types::beacon::events::HeadEvent, signers};
-use tokio::sync::mpsc;
-
-use bolt_sidecar::{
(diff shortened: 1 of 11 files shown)


thedevbirb (approved):
Thanks for the refactor! Very needed :)

namn-grg (commented):
Great work ser! Love this refactor!

PR #185, opened 2024-08-06
chore(mev-boost-relay): set SSZ max constraints per slot to the defau…

### LINKED ISSUE #61: Decide the value of the constant `MAX_CONSTRAINTS_PER_SLOT`
In order to encode constraints using SSZ we need to provide a maximum length to list. Regarding constraints, I decided for now to just set it to `MAX_CONSTRAINTS_PER_SLOT = 256` but this is up to discussion whether this value is reasonable or not.

Other than changing the value in the codebase and in the SSZ encoding functions, we should also define this constant value in the specs website

merklefruit:
…lt max transactions per block

closes #61

### DIFF
--- a/mev-boost-relay/services/api/constraints.go
+++ b/mev-boost-relay/services/api/constraints.go
@@ -34,7 +34,7 @@ type ConstraintsMessage struct {
 }
 
 type Constraint struct {
-	Tx    Transaction `ssz-max:"1048576" json:"tx"`
+	Tx    Transaction `ssz-max:"1073741824" json:"tx"`
 	Index *Index      `json:"index"`
 }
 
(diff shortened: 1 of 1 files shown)


mev-boost-relay/services/api/constraints.go
```diff
@@ -25,15 +30,15 @@ type SignedConstraints struct {
 type ConstraintsMessage struct {
 	ValidatorIndex uint64        `json:"validator_index"`
 	Slot           uint64        `json:"slot"`
-	Constraints    []*Constraint `ssz-max:"256" json:"constraints"`
+	Constraints    []*Constraint `ssz-max:"1048576" json:"constraints"`
 }
 
 type Constraint struct {
-	Tx    Transaction `ssz-max:"1048576" json:"tx"`
+	Tx    Transaction `ssz-max:"1073741824" json:"tx"`
```
merklefruit, line 37: Note: this value was wrong, however we only use JSON for this type currently.
We should either remove SSZ definitions for it or start using it.

mempirate (approved):
Lgtm!

PR #186, opened 2024-08-06
wip: perform diffs of all flashbots packages

merklefruit:
This will create `.diff` files for internal use to assess changes with upstream flashbots packages in a simple way.
This doesn't include full `fork.yaml` definitions needed to build the forkdiff websites yet, but it's a first step.

*I hate bash*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *_dump.log
+forkdiff/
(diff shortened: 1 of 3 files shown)


merklefruit:
We can easily skip this one as we already have forkdiffs online now.

PR #187, opened 2024-08-06
chore(builder,relay): remove auth when subscribing for constraints

### LINKED ISSUE #41: Specify authorization schema for builders
Builders need to authenticate when subscribing to SSE events emitted from relays. Right now in the codebase we're using a string like this:
```
"BOLT \"<signature>\",{\"publicKey\":\"<builder_pub_key>\",\"slot\":<slot>}"
```
where `signature` refers to the JSON-stringified object appearing the in the authentication message.
However, we should discuss more about it and update the specs.

The rationale for this schema is to have a simple one that included the builder public key, a nonce-like number to avoid replay attacks and a signature

merklefruit:
closes #41 at the latest comment update.

### DIFF
--- a/mev-boost-relay/services/api/service.go
+++ b/mev-boost-relay/services/api/service.go
@@ -3047,7 +3047,8 @@ func (api *RelayAPI) handleSubscribeConstraints(w http.ResponseWriter, req *http
 			return
 		case <-ticker.C:
 			// Send a keepalive to the client
-			fmt.Fprint(w, ": keepalive\n\n")
+			// NOTE: the length of the message is intentional, do not make it shorter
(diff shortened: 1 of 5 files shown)


mev-boost-relay/services/api/service.go
```diff
@@ -3065,7 +3047,7 @@ func (api *RelayAPI) handleSubscribeConstraints(w http.ResponseWriter, req *http
 			return
 		case <-ticker.C:
 			// Send a keepalive to the client
-			fmt.Fprint(w, ": keepaliveeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee\n\n")
```
mempirate, line 3068: NGL I think there was an actual reason for making it this long, something related to buffering 😩

merklefruit: Lmao, ok reverting 🚀

mempirate (approved):

PR #188, opened 2024-08-07
fix: commitment deadline in case of missed block

### LINKED ISSUE #180: Bug: missed block fails to kick off sidecar round
## Context
To initiate a sidecar "round" i.e. submit constraints, we rely on a new block stream. However, when no block is proposed (missed), we fail to initiate the round. We should look at a more fault-tolerant way of "keeping time", perhaps subscribe to the beacon block stream instead.

## Game plan
- Is there any events stream on the [beacon-API](https://ethereum.github.io/beacon-APIs/#/Events/eventstream) that ticks no matter what? (`head` or `block`)
- If not we should rely on the system clock. The most important thing is that the logic related to timing actually fires.
- Note that we cannot use this system clock to advance the execution / consensus state, because in most cases the block for that slot will not have arrived yet. This needs to be decoupled somehow.

namn-grg:
Fixes #180 

## Summary by CodeRabbit

- **New Features**
	- Enhanced consensus management with the addition of a `consensus_clock` for improved time representation related to consensus slots.
	- New method to retrieve the cached beacon genesis timestamp from the `HeadTracker`.

- **Bug Fixes**
	- Improved resilience in fetching the beacon genesis timestamp with a retry mechanism.

- **Refactor**
	- Simplified `ConsensusState` by removing the `header` field and renaming `update_head` to `update_slot` for clearer slot management.
	- Standardized retry logic for subscribing to new head events in the `HeadTracker`.

- **Tests**
	- Added test cases to verify the correct behavior of the updated slot management functionality.

### DIFF
--- a/bolt-sidecar/src/driver.rs
+++ b/bolt-sidecar/src/driver.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use std::time::{Duration, Instant};
 
 use alloy::{
@@ -5,6 +6,11 @@ use alloy::{
     signers::{local::PrivateKeySigner, Signer as SignerECDSA},
 };
 use beacon_api_client::mainnet::Client as BeaconClient;
+use ethereum_consensus::{
+    clock::{from_system_time, Clock, SystemTimeProvider},
+    phase0::mainnet::SLOTS_PER_EPOCH,
+};
+use futures::StreamExt;
 use tokio::sync::mpsc;
 use tracing::{debug, error, info, warn};
 
@@ -24,7 +30,6 @@ use crate::{
 };
 
 /// The driver for the sidecar, responsible for managing the main event loop.
-#[derive(Debug)]
 pub struct SidecarDriver<C, BLS, ECDSA> {
     head_tracker: HeadTracker,
     execution: ExecutionState<C>,
@@ -35,6 +40,23 @@ pub struct SidecarDriver<C, BLS, ECDSA> {
     mevboost_client: MevBoostClient,
     api_events_rx: mpsc::Receiver<CommitmentEvent>,
     payload_requests_rx: mpsc::Receiver<FetchPayloadRequest>,
+    consensus_clock: Clock<SystemTimeProvider>,
+}
+
+impl fmt::Debug for SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SidecarDriver")
+            .field("head_tracker", &self.head_tracker)
+            .field("execution", &self.execution)
+            .field("consensus", &self.consensus)
+            .field("constraint_signer", &self.constraint_signer)
+            .field("commitment_signer", &self.commitment_signer)
+            .field("local_builder", &self.local_builder)
+            .field("mevboost_client", &self.mevboost_client)
+            .field("api_events_rx", &self.api_events_rx)
+            .field("payload_requests_rx", &self.payload_requests_rx)
+            .finish()
+    }
 }
 
 impl SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/driver.rs
```diff
@@ -74,17 +96,24 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
...
+        // but using beacon node events is easier to work on a custom devnet for now
+        // (as we don't need to specify genesis time and slot duration)
+        let head_tracker = HeadTracker::start(beacon_client.clone());
+
+        // Initialize the consensus clock.
+        let consensus_clock = from_system_time(
+            head_tracker.beacon_genesis_timestamp(),
+            cfg.chain.slot_time(),
+            SLOTS_PER_EPOCH,
+        );
```
merklefruit: reminder: this is not great because head tracker initializes the genesis timestamp to 0.
We should add a fn to simply async fetch that value instead for safety

This can be done in another issue

bolt-sidecar/src/driver.rs
```diff
@@ -110,6 +139,7 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
             mevboost_client,
             api_events_rx,
             payload_requests_rx,
+            consensus_clock,
```
mempirate: We can already turn it into a stream here!

bolt-sidecar/src/state/consensus.rs
```diff
@@ -117,22 +114,17 @@ impl ConsensusState {
     }
 
     /// Update the latest head and fetch the relevant data from the beacon chain.
-    pub async fn update_head(&mut self, head: u64) -> Result<(), ConsensusError> {
+    pub async fn update_slot(&mut self, head: u64) -> Result<(), ConsensusError> {
```
mempirate: Rename param `head` -> `slot` for consistency

bolt-sidecar/src/driver.rs
```diff
@@ -35,6 +40,23 @@ pub struct SidecarDriver<C, BLS, ECDSA> {
     mevboost_client: MevBoostClient,
     api_events_rx: mpsc::Receiver<CommitmentEvent>,
     payload_requests_rx: mpsc::Receiver<FetchPayloadRequest>,
+    consensus_clock: Clock<SystemTimeProvider>,
```
merklefruit: if we only need the SlotStream from the clock, I would prefer to keep that in the struct instead of the Clock.
It should be of type `SlotStream<something>` but if it's an issue let me know!

bolt-sidecar/src/driver.rs
```diff
@@ -74,17 +96,24 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
         let (api_events_tx, api_events_rx) = mpsc::channel(1024);
         CommitmentsApiServer::new(api_addr).run(api_events_tx).await;
 
+        // TODO: this can be replaced with ethereum_consensus::clock::from_system_time()
+        // but using beacon node events is easier to work on a custom devnet for now
+        // (as we don't need to specify genesis time and slot duration)
+        let head_tracker = HeadTracker::start(beacon_client.clone());
+
+        // Initialize the consensus clock.
+        let consensus_clock = from_system_time(
```
merklefruit: can we leave the package name for clarity here? e.g. `clock::from_system_time()` otherwise it's a bit unclear imo

mempirate (changes requested):
Nice one, some smol nits

bolt-sidecar/src/driver.rs
```diff
@@ -132,6 +164,11 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
                 Some(payload_request) = self.payload_requests_rx.recv() => {
                     self.handle_fetch_payload_request(payload_request);
                 }
+                Some(slot) = slot_stream.next() => {
+                    if let Err(e) = self.consensus.update_slot(slot).await {
+                        error!(err = ?e, "Failed to update consensus state slot");
+                    }
```
merklefruit: Awesome!

merklefruit (changes requested):
Very nice work! Just left some minor comments

namn-grg:
@coderabbitai review

namn-grg:
Thanks for the speedy review! Addressed it

bolt-sidecar/src/driver.rs
namn-grg: I added a comment for this.

but I don't understand how the soln for this should be.

```
let genesis_time = loop {
                    match beacon_client.get_genesis_details().await {
                        Ok(genesis_info) => break genesis_info.genesis_time,
                        Err(err) => {
                            warn!(?err, "failed to get genesis details");
                            sleep(RETRY_DELAY).await;
                            continue;
                        }
                    }
};
```

I am fetching the value like this

merklefruit: on second thought, we should really have this value in `ChainConfig` anyway - for now it's fine though 👍

merklefruit (approved):
LGTM!

PR #194, opened 2024-08-09
fix: build local payload with timestamp from beacon clock

merklefruit:
This PR fixes the edge case in which the previous slot was missed by the proposer before us, and we would
use the wrong block timestamp. Now we use the timestamp according to the beacon proposal slot instead.

This also contains minor refactoring to the local payload fetcher to be a little cleaner.

### DIFF
--- a/bolt-sidecar/src/builder/payload_builder.rs
+++ b/bolt-sidecar/src/builder/payload_builder.rs
@@ -48,12 +48,13 @@ pub struct FallbackPayloadBuilder {
     beacon_api_client: BeaconClient,
     execution_rpc_client: RpcClient,
     engine_hinter: EngineHinter,
-    slot_time_in_seconds: u64,
+    slot_time: u64,
+    genesis_time: u64,
 }
 
 impl FallbackPayloadBuilder {
     /// Create a new fallback payload builder
-    pub fn new(config: &Config) -> Self {
+    pub fn new(config: &Config, beacon_api_client: BeaconClient, genesis_time: u64) -> Self {
         let engine_hinter = EngineHinter {
             client: reqwest::Client::new(),
             jwt_hex: config.jwt_hex.to_string(),
@@ -64,9 +65,10 @@ impl FallbackPayloadBuilder {
             engine_hinter,
             extra_data: DEFAULT_EXTRA_DATA.into(),
             fee_recipient: config.fee_recipient,
-            beacon_api_client: BeaconClient::new(config.beacon_api_url.clone()),
             execution_rpc_client: RpcClient::new(config.execution_api_url.clone()),
-            slot_time_in_seconds: config.chain.slot_time(),
+            slot_time: config.chain.slot_time(),
+            genesis_time,
+            beacon_api_client,
         }
     }
 }
@@ -86,7 +88,7 @@ struct Context {
     transactions_root: B256,
     withdrawals_root: B256,
     parent_beacon_block_root: B256,
-    slot_time_in_seconds: u64,
+    block_timestamp: u64,
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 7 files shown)


bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -243,6 +227,43 @@ impl FallbackPayloadBuilder {
...
+    }
+
+    /// Fetch the expected withdrawals for the given slot from the beacon chain.
+    async fn get_withdrawals_for_slot(
+        &self,
+        slot: u64,
+    ) -> Result<Vec<reth_primitives::Withdrawal>, BuilderError> {
+        Ok(self
+            .beacon_api_client
+            .get_expected_withdrawals(StateId::Head, Some(slot))
```
merklefruit: not 100% sure if it's better to provide the target `slot` here or leave `None` to make the CL use the latest head by default. Would love input on this

bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -167,6 +146,11 @@ impl FallbackPayloadBuilder {
         let blob_gas_used =
             transactions.iter().fold(0, |acc, tx| acc + tx.blob_gas_used().unwrap_or_default());
 
+        // We must calculate the next block timestamp manually rather than rely on the
+        // previous execution block, to cover the edge case where any previous slots have
+        // been missed by the proposers immediately before us.
+        let block_timestamp = self.genesis_time + (target_slot * self.slot_time);
```
merklefruit, line 152: this should be the new timestamp because target_slot is the one we are expected to propose at.

bolt-sidecar/src/builder/payload_builder.rs
mempirate: As it seemed to be working previously I wouldn't change it

bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -243,6 +227,42 @@ impl FallbackPayloadBuilder {
...
+            .pointer("/data/randao")
+            .and_then(|value| value.as_str())
+            .map(|value| B256::from_hex(value).map_err(BuilderError::Hex))
+            .ok_or_else(|| BuilderError::Custom("Failed to fetch prev_randao".to_string()))?
+    }
+
+    /// Fetch the expected withdrawals for the given slot from the beacon chain.
+    async fn get_expected_withdrawals_at_head(
+        &self,
+    ) -> Result<Vec<reth_primitives::Withdrawal>, BuilderError> {
```
namn-grg:
```suggestion
    ) -> Result<Vec<Withdrawal>, BuilderError> {
```
nit - better to import at the top? :)

bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -243,6 +227,42 @@ impl FallbackPayloadBuilder {
             i += 1;
         }
     }
+
+    /// Fetch the previous RANDAO value from the beacon chain.
+    ///
+    /// NOTE: for some reason, using the ApiResult from `beacon_api_client` doesn't work, so
+    /// we are making a direct request to the beacon client endpoint.
```
namn-grg, line 234: hmm just curious to know some more detail

merklefruit: I think it has to do with the type that the `beacon_api_client` crate expects to deserialize from the result of that API call - maybe it has changed or it's not updated to what Lighthouse is responding, so it fails to deserialize.

This is probably a bug in the beacon_api_client crate - but I haven't investigated

namn-grg (approved):

namn-grg:
Great catch ser!

mempirate (approved):
LGTM!

Issue #201, opened 2024-08-22
Commit-Boost integration


mempirate:
## Context
We'd like to support running Bolt as a commit-boost module. For Bolt, this means 2 things:
- If run with commit-boost, signatures should be requested from commit-boost. There are 2 types of signatures: BLS for authentication (what constraints should be signed with), and ECDSA for binding commitments. commit-boost has an open issue for ECDSA: https://github.com/Commit-Boost/commit-boost-client/pull/87. We can provide a flag for `--signer-api` which would indicate to the process that it is running alongside commit-boost.
- We also need to modify the regular PBS module to support the constraints-API. An example can be found here: https://github.com/eserilev/il-boost

Commit-boost documentation: https://commit-boost.github.io/commit-boost-client

mempirate:
Closing this as the final touches will be added when testing the whole flow.

PR #203, opened 2024-08-28
feat(sidecar): commit-boost integration

namn-grg:
partially solves #201 

Note - This is not e2e tested with commit-boost

## Summary by CodeRabbit

- **New Features**
	- Introduced new configuration options for commit boosting, including an API endpoint and JWT for authentication.
	- Added a new client type, `CommitBoostSigner`, to the public interface.
	
- **Improvements**
	- Enhanced BLS signing functionality with support for different signer types and improved type safety.
	- Updated the `SidecarDriver` to streamline the signing process and support asynchronous operations.
	- Improved the flexibility and security of the configuration management system.

- **Bug Fixes**
	- Simplified error handling in the signature request process and enhanced the clarity of the `commit_and_sign` method.

- **Chores**
	- Removed unnecessary dependencies to improve project requirements.


bolt-sidecar/src/config/mod.rs
```diff
@@ -180,8 +183,11 @@ impl TryFrom<Opts> for Config {
             config.limits.max_committed_gas_per_slot = max_committed_gas;
         }
 
-        config.commit_boost_url =
-            opts.signing.commit_boost_url.as_ref().map(|url| Url::parse(url)).transpose()?;
+        config.commit_boost_address = opts
+            .signing
+            .commit_boost_url
+            .as_ref()
+            .map(|url| url.strip_prefix("http://").unwrap_or(url).to_string());
```
namn-grg: We will have to remove the "http://" beacuse apparently the commit-boost new impl takes only the _address_ https://github.com/Commit-Boost/commit-boost-client/blob/828b10811f125bcc8ab5057df033aa29286d6130/crates/common/src/commit/client.rs#L31

I thought its cleaner to take the url as input, as in all the env variables and just remove the http here.

bolt-sidecar/src/crypto/bls.rs
```diff
@@ -49,6 +50,25 @@ pub trait SignerBLSAsync: Send + Sync {
...
 }
 
+/// Enum to represent the two possible signer types.
+#[derive(Debug)]
+pub enum BlsSignerType {
+    /// A BLS signer that uses private key to sign.
+    PrivateKey(Signer),
+    /// A BLS signer that request signature from commit-boost.
+    CommitBoost(CommitBoostClient),
+}
```
namn-grg: Not fully sure if this is the best way to support both, I tried doing some other ways but this seemed to me the most cleaner

namn-grg:
@coderabbitai review

bolt-sidecar/src/config/mod.rs
mempirate: To cover for cases where users provide an https address as well, we can parse the string into a `reqwest::Url` and then extract the `SocketAddr`

bolt-sidecar/src/driver.rs
```diff
@@ -78,11 +89,11 @@ impl SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
     }
 }
 
-impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS, ECDSA> {
+impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
```
merklefruit: I think we can still have the generic here, by changing the bind to the bls async trait:

```rust
impl<C: StateFetcher, BLS: SignerBLSAsync, ECDSA: SignerECDSA> SidecarDriver<C, BLS, ECDSA> {
```

then we shouldn't need the `BlsSignerType`

bolt-sidecar/src/driver.rs
```diff
@@ -60,15 +63,23 @@ impl fmt::Debug for SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
...
+        let constraint_signer = if let Some(private_key) = cfg.private_key.clone() {
+            BlsSignerType::PrivateKey(BlsSigner::new(private_key))
+        } else {
+            let commit_boost_client = CommitBoostClient::new(
+                cfg.commit_boost_address.clone().expect("CommitBoost URL must be provided"),
+                &cfg.commit_boost_jwt_hex.clone().expect("CommitBoost JWT must be provided"),
+            )
+            .await?;
+            BlsSignerType::CommitBoost(commit_boost_client)
+        };
```
merklefruit: It would be nice to refactor the signing config (`SigningOpts`) to avoid unwrapping CLI options here.
can be done in a separate PR.

bolt-sidecar/src/driver.rs
```diff
@@ -60,15 +63,23 @@ impl fmt::Debug for SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
     }
 }
 
-impl SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
+impl SidecarDriver<StateClient, PrivateKeySigner> {
```
merklefruit: to keep the BLS generic trait in the SidecarDriver, we will need to have 2 separate associated methods `new()`.
One for the private key signer and one for commit-boost. Something like this:

```rust
impl SidecarDriver<StateClient, LocalBlsSigner, PrivateKeySigner> {
	pub fn new(cfg: Config) -> eyre::Result<Self> {
		// ...
		Self::from_components(...)
	}
}

impl SidecarDriver<StateClient, CommitBoostSigner, PrivateKeySigner> {
	pub fn new(cfg:Config) -> eyre::Result<Self> {
		// ...
		Self::from_components(...)
	}
}
```

bolt-sidecar/src/crypto/bls.rs
```diff
@@ -36,7 +38,6 @@ pub trait SignableBLS {
 }
 
 /// A generic signing trait to generate BLS signatures.
-#[allow(dead_code)]
 pub trait SignerBLS {
     /// Sign the given data and return the signature.
     fn sign(&self, data: &[u8]) -> eyre::Result<BLSSig>;
```
mempirate: Since BLS signers should only sign hash tree roots, perhaps we can change the param (here and in `SignerBLSAsync` to `root: [u8; 32]`?

mempirate (changes requested):
Nice job, take a look at Nico's comment to use a BLS signer trait instead of an enum

bolt-sidecar/src/crypto/bls.rs
namn-grg: yes makes sense done!

bolt-sidecar/src/driver.rs
namn-grg: thanks good idea! done

bolt-sidecar/bin/sidecar.rs
```diff
@@ -13,10 +13,20 @@ async fn main() -> Result<()> {
...
+            Err(err) => bail!("Failed to initialize the sidecar driver: {:?}", err),
+        }
+    } else {
+        match SidecarDriver::with_commit_boost_signer(config).await {
+            Ok(driver) => driver.run_forever().await,
+            Err(err) => {
+                bail!("Failed to initialize the sidecar driver with commit boost: {:?}", err)
+            }
+        }
+    }
```
merklefruit, line 31: lovely

bolt-sidecar/src/config/mod.rs
```diff
@@ -180,8 +183,13 @@ impl TryFrom<Opts> for Config {
             config.limits.max_committed_gas_per_slot = max_committed_gas;
         }
 
-        config.commit_boost_url =
-            opts.signing.commit_boost_url.as_ref().map(|url| Url::parse(url)).transpose()?;
+        if let Some(commit_boost_url) = &opts.signing.commit_boost_url {
+            if let Ok(url) = Url::parse(commit_boost_url) {
+                if let Ok(socket_addrs) = url.socket_addrs(|| None) {
```
merklefruit, line 200: what's with the `|| None` closure? just curious

bolt-sidecar/src/driver.rs
```diff
@@ -44,7 +44,7 @@ pub struct SidecarDriver<C, BLS, ECDSA> {
     slot_stream: SlotStream<SystemTimeProvider>,
 }
 
-impl fmt::Debug for SidecarDriver<StateClient, BlsSigner, PrivateKeySigner> {
+impl fmt::Debug for SidecarDriver<StateClient, Box<dyn SignerBLSAsync>, PrivateKeySigner> {
```
merklefruit:
```suggestion
impl<B: SignerBLSAsync> fmt::Debug for SidecarDriver<StateClient, B, PrivateKeySigner> {
```

bolt-sidecar/src/config/mod.rs
namn-grg: This is for matching the url schema, since we dont have any here so have to specify None. 

> If the URL has the default port number of a scheme that is unknown to this library, default_port_number provides an opportunity to provide the actual port number. In non-example code this should be implemented either simply as || None, or by matching on the URL’s .scheme().

merklefruit: got it ty

bolt-sidecar/src/driver.rs
namn-grg: thanks for pointing out! :)

namn-grg:
@coderabbitai review

merklefruit (approved):
Great work!

bolt-sidecar/src/config/signing.rs
```diff
@@ -13,4 +13,7 @@ pub struct SigningOpts {
     /// URL for the commit-boost sidecar
     #[clap(long, env = "BOLT_SIDECAR_COMMIT_BOOST_URL", conflicts_with("private_key"))]
```
mempirate: Let's rename env to `BOLT_SIDECAR_CB_SIGNER_URL` (to distinguish PBS module from commit boost)

And add `requires("commit_boost_jwt_hex")`

bolt-sidecar/src/crypto/ecdsa.rs
```diff
@@ -51,19 +54,38 @@ impl ECDSASigner {
     }
 }
 
+/// A generic signing trait to generate ECDSA signatures.
+#[async_trait::async_trait]
+pub trait SignerECDSAAsync: Send + Sync + Debug {
```
mempirate: If we never use the synchronous trait, let's drop that and rename this to `SignerECDSA` (same with `SignerBLS`)

mempirate (changes requested):
Great work. Some small nits and then we can merge!

bolt-sidecar/src/crypto/ecdsa.rs
namn-grg: good observation! made the changes

merklefruit (approved):

namn-grg:
Just working on a mock test

bolt-sidecar/src/client/commit_boost.rs
```diff
@@ -101,3 +141,41 @@ impl SignerECDSA for CommitBoostSigner {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rand::Rng;
+
+    #[tokio::test]
+    async fn test_bls_commit_boost_signer() {
+        let signer =
+            CommitBoostSigner::new("http://localhost:19551".to_string(), "jwt_hex").await.unwrap();
```
mempirate: This will fail in CI, we should use a pattern similar to the others where we read env variables to get the URL, and if not provided, skip the tests.

namn-grg: mb, updated!

PR #204, opened 2024-08-28
Telemetry & Metrics

### LINKED ISSUE #116: Sidecar: Telemetry & Metrics
```[tasklist]
### Bolt-sidecar telemetry
- [x] Review the level of existing traces, and change or remove any unnecessary ones
- [x] Add the option to instrument a Prometheus listener to collect useful metrics
- [x] Add meaningful metrics to the sidecar, including health-checks, alerts, and important events
- [x] Infra: create a default Grafana dashboard to visualize the sidecar status
- [ ] (optional) Add the option to instrument a log collector such as a remote Loki instance
- [ ] (optional) Add the option to log to a file in JSON to support external log collectors like Promtail
```

thedevbirb:
Closes #116. 
Introduces telemetry and metrics into the Bolt sidecar, which are enabled using the `metrics` flag, by spinning up a Prometheus server and basic Grafana dashboard, which are available also for the devnet (see https://github.com/chainbound/ethereum-package/pull/7).

The integration with Promtail and Loki will be addressed in a future PR if needed.

NOTE:
- the docker compose setup using the `telemetry.compose.yml` is currently non-tested on Helder. Main testing has been done with devnet data.
- needs to be merged along with https://github.com/chainbound/ethereum-package/pull/7

Lastly, most of the code is heavily inspired by the work of @merklefruit and @namn-grg on https://github.com/chainbound/bolt-rpc, so thanks a lot for the guideline!

image

### DIFF
--- a/bolt-sidecar/src/config/mod.rs
+++ b/bolt-sidecar/src/config/mod.rs
@@ -6,7 +6,6 @@ use clap::Parser;
 use eyre::{bail, eyre, Report, Result};
 use reqwest::Url;
 use std::num::NonZero;
-use telemetry::TelemetryOpts;
 use tracing::info;
 
 use crate::crypto::bls::random_bls_secret;
@@ -21,6 +20,7 @@ pub mod signing;
 pub use signing::SigningOpts;
 
 pub mod telemetry;
+use telemetry::TelemetryOpts;
 
 /// Default port for the JSON-RPC server exposed by the sidecar.
 pub const DEFAULT_RPC_PORT: u16 = 8000;
@@ -123,7 +123,7 @@ pub struct Config {
     /// Metrics port
     pub metrics_port: u16,
     /// Toggle for metrics
-    pub use_metrics: bool,
+    pub metrics: bool,
 }
 
 impl Default for Config {
@@ -144,7 +144,7 @@ impl Default for Config {
             validator_indexes: ValidatorIndexes::default(),
             chain: ChainConfig::default(),
             metrics_port: 0,
-            use_metrics: true,
+            metrics: true,
         }
     }
 }
@@ -241,7 +241,7 @@ impl TryFrom<Opts> for Config {
 
         config.chain = opts.chain;
         config.metrics_port = opts.telemetry.metrics_port;
-        config.use_metrics = opts.telemetry.use_metrics;
+        config.metrics = opts.telemetry.metrics;
 
         Ok(config)
     }
--- a/bolt-sidecar/src/config/mod.rs
+++ b/bolt-sidecar/src/config/mod.rs
@@ -120,6 +120,8 @@ pub struct Config {
     pub builder_private_key: SecretKey,
     /// The chain on which the sidecar is running
     pub chain: ChainConfig,
+    /// Metrics port
+    pub metrics_port: u16,
 }
 
 impl Default for Config {
@@ -139,6 +141,7 @@ impl Default for Config {
             limits: Limits::default(),
             validator_indexes: ValidatorIndexes::default(),
             chain: ChainConfig::default(),
+            metrics_port: 0,
         }
     }
 }
@@ -234,6 +237,7 @@ impl TryFrom<Opts> for Config {
         config.validator_indexes = opts.validator_indexes;
 
         config.chain = opts.chain;
+        config.metrics_port = opts.telemetry.metrics_port;
 
         Ok(config)
     }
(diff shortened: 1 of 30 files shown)


merklefruit:
can you add some docs related to telemetry in the sidecar's readme?

bolt-sidecar/src/config/mod.rs
```diff
@@ -6,6 +6,7 @@ use clap::Parser;
 use eyre::{bail, eyre, Report, Result};
 use reqwest::Url;
 use std::num::NonZero;
+use telemetry::TelemetryOpts;
```
merklefruit: can you move this below `pub mod telemetry`?

bolt-sidecar/src/config/mod.rs
```diff
@@ -114,6 +120,10 @@ pub struct Config {
     pub builder_private_key: SecretKey,
     /// The chain on which the sidecar is running
     pub chain: ChainConfig,
+    /// Metrics port
+    pub metrics_port: u16,
+    /// Toggle for metrics
+    pub use_metrics: bool,
```
merklefruit: the more "idiomatic" way to start metrics is `--metrics` that e.g. Geth and Reth use.
Reference: https://github.com/paradigmxyz/reth/blob/019ec727a2e078fd1517d5b04f39f6713d3bf0d7/crates/cli/commands/src/node.rs#L46

bolt-sidecar/src/state/execution.rs
```diff
@@ -86,6 +86,35 @@ impl ValidationError {
...
+            }
+            ValidationError::MaxCommittedGasReachedForSlot(_, _) => {
+                "max_committed_gas_reached_for_slot"
+            }
+            ValidationError::Signature(_) => "signature",
+            ValidationError::RecoverSigner => "recover_signer",
+            ValidationError::ChainIdMismatch => "chain_id_mismatch",
+            ValidationError::Internal(_) => "internal",
+        }
+    }
```
merklefruit, line 120: I wonder if there is a better way to do this, maybe with `#[serde(rename_all = "snake_case")]` or something

bolt-sidecar/src/telemetry/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    // 2. Initialize metrics recorder and start the Prometheus server
+    let prometheus_addr = SocketAddr::from(([0, 0, 0, 0], metrics_port));
+    let builder = PrometheusBuilder::new().with_http_listener(prometheus_addr);
+
+    if let Err(e) = builder.install() {
+        bail!("failed to install Prometheus recorder: {:?}", e);
+    } else {
+        info!("Telemetry initialized. Serving Prometheus metrics at: http://{}", prometheus_addr);
+    }
```
merklefruit: should be made optional if `use_metrics` is true, no?

merklefruit (changes requested):

merklefruit (approved):
LGTM!

bolt-sidecar/src/config/mod.rs
```diff
@@ -114,6 +120,10 @@ pub struct Config {
     pub builder_private_key: SecretKey,
     /// The chain on which the sidecar is running
     pub chain: ChainConfig,
+    /// Metrics port
+    pub metrics_port: u16,
+    /// Toggle for metrics
+    pub metrics: bool,
```
mempirate: Can we make this `--disable-metrics` instead?

namn-grg (approved):
Great PR! Love it awesome!

bolt-sidecar/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,153 @@
...
+
+#[allow(missing_docs)]
+impl ApiMetrics {
+    pub fn start() {
+        ApiMetricType::describe_all();
+    }
+
+    /// Counters ----------------------------------------------------------------
+
+    pub fn increment_total_http_requests(method: String, path: String, status: String) {
```
merklefruit, line 60: quick question: does this work with &str or only String as arguments? I think it would be more ergonomic to use with slices if that's doable. If not, that's ok

thedevbirb: No in general labels must be owned values. Since when this function is called owned strings are available I thought this function signature would be better instead of using &str, even if the latter is more idiomatic

merklefruit: yup that's perfectly fine, thanks

PR #207, opened 2024-09-02
Bolt-boost module

### LINKED ISSUE #90: Implement hash tree root for `ConstraintsMessage`
In our fork of Flashbots' relay we added an endpoint to submit signed constraints (https://chainbound.github.io/bolt-docs/api/constraints-api#relayv1builderconstraints), with the underlying message of type `ConstraintsMessage`. 
Right now, we verify signatures using the SSZ encoding as digest; we should use the hash tree root of such container instead, as done with other containers in the consensus specs.

```[tasklist]
### Tasks
- [ ] Implement `fastssz`'s `HashTreeRoot` interface for `ConstraintsMessage` (with tests)
- [ ] Modify `handleSubmitConstraints` signature checks by using the hash tree root as a digest
```

mempirate:
## Context
This PR implements a new commit-boost module with the constraints-API.

Closes #90


bolt-boost/src/constraints.rs
```diff
@@ -0,0 +1,87 @@
...
+    sync::{Arc, RwLock},
+};
+use tracing::error;
+
+use super::types::{ConstraintsMessage, ConstraintsWithProofData};
+
+/// A concurrent cache of constraints.
+#[derive(Clone, Default, Debug)]
+pub struct ConstraintsCache {
+    cache: Arc<RwLock<HashMap<u64, Vec<ConstraintsWithProofData>>>>,
```
merklefruit, line 15: nit: can we use `parking_lot`'s rwlock?

bolt-boost/src/proofs.rs
```diff
@@ -0,0 +1,226 @@
...
+    use alloy::{
+        hex::FromHex,
+        primitives::{Bytes, B256},
+    };
+    use ssz_rs::{HashTreeRoot, List, PathElement, Prove};
+
+    use crate::testutil::*;
+
+    // #[test]
+    // fn test_single_multiproof() {
```
merklefruit, line 91: is this test missing testdata or breaking for other reasons?

bolt-boost/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+futures = "0.3"
+async-trait = "0.1.82"
+eyre = "0.6.12"
+thiserror = "1.0.63"
+reqwest = "0.12.7"
+
+# crypto
+ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs", features = [
+    "sha2-asm",
+] }
```
thedevbirb, line 20: Can you pin a version or commit?

bolt-boost/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+    "rpc-types-beacon",
+    "rpc-types-engine",
+] }
+
+# commit-boost
+commit-boost = { git = "https://github.com/commit-boost/commit-boost-client" }
+cb-common = { git = "https://github.com/commit-boost/commit-boost-client" }
+
+# other
+types = { git = "https://github.com/sigp/lighthouse" }
```
thedevbirb, line 45: same for those three

bolt-boost/src/constraints.rs
```diff
@@ -0,0 +1,87 @@
...
+            cache: Default::default(),
+        }
+    }
+
+    /// Checks if the constraints for the given slot conflict with the existing constraints.
+    /// Will check for:
+    /// - Multiple ToB constraints per slot
+    /// - Duplicates of the same transaction per slot
+    pub fn conflicts_with(&self, slot: &u64, constraints: &ConstraintsMessage) -> bool {
+        if let Some(saved_constraints) = self.cache.read().unwrap().get(slot) {
```
thedevbirb, line 27: A nit but can we turn the `unwrap` into an `expect` with a reason?

bolt-boost/src/constraints.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        false
+    }
+
+    /// Inserts the constraints for the given slot. Also decodes the raw transactions to save their
+    /// transaction hashes and hash tree roots for later use. Will first check for conflicts, and return
+    /// false if there are any.
+    ///
+    /// TODO: return Result instead of bool
```
thedevbirb, line 55: TODO left: do you want to address it in a future PR?

bolt-boost/Cargo.toml
thedevbirb: Can you pin the version or commit?

bolt-boost/Cargo.toml
thedevbirb: Same for these three dependencies

bolt-boost/src/constraints.rs
thedevbirb: It's a nit but can you turn the `unwrap` into an `expect` with a reason?

bolt-boost/src/constraints.rs
thedevbirb: TODO left here: do you want to address this in a future PR?

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,582 @@
...
+            max_timeout_ms,
+        ));
+    }
+
+    let results = join_all(handles).await;
+    let mut relay_bids = Vec::with_capacity(relays.len());
+    let mut hash_to_proofs = HashMap::new();
+
+    // Get and remove the constraints for this slot
+    let constraints = state.data.constraints.remove(params.slot);
```
merklefruit: not checking this Option will result in panicking if there are no constraints for this slot. is this safely handled?

bolt-boost/src/proofs.rs
```diff
@@ -0,0 +1,226 @@
...
+            if found {
+                break;
+            }
+        }
+
+        // If the hash is not found in the constraints cache, return an error
+        if !found {
+            return Err(ProofError::MissingHash(*hash));
+        }
+    }
```
thedevbirb, line 65: Can you a comment on top of this block to explain what it does? Even if quite clear after a careful read, it is useful when skimming over the code

bolt-boost/src/proofs.rs
```diff
@@ -0,0 +1,226 @@
...
+        hex::FromHex,
+        primitives::{Bytes, B256},
+    };
+    use ssz_rs::{HashTreeRoot, List, PathElement, Prove};
+
+    use crate::testutil::*;
+
+    // #[test]
+    // fn test_single_multiproof() {
+    //     let (root, transactions) = read_test_transactions();
```
thedevbirb, line 92: Is this intended to be left as a comment?

bolt-sidecar/Cargo.toml
```diff
@@ -64,11 +65,12 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter", "fmt"] }
...
 metrics = "0.23"
 metrics-exporter-prometheus = { version = "0.15.3", features = [
-  "http-listener",
+    "http-listener",
 ] }
 
 # commit-boost
-cb-common = { git = "https://github.com/Commit-Boost/commit-boost-client", rev = "f1e6ff1" }
+commit-boost = { git = "https://github.com/Commit-Boost/commit-boost-client" }
+cb-common = { git = "https://github.com/Commit-Boost/commit-boost-client" }
```
thedevbirb, line 73: Same here about pinning versions

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,582 @@
...
+
+                relay_bids.push(vanilla_response)
+            }
+            Ok(_) => {}
+            Err(err) if err.is_timeout() => error!(err = "Timed Out", relay_id),
+            Err(err) => error!(?err, relay_id),
+        }
+    }
+
+    let header = state.add_bids(params.slot, relay_bids);
```
merklefruit: it's not intuitive that this function returns the best bid in the current cache. can we make it more explicit by renaming it to `best_header` or `best_relay_bid`?

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,582 @@
...
+            header: h.data,
+        },
+        version: h.version,
+    });
+
+    if let Some(header_with_proofs) = header_with_proofs {
+        Ok((StatusCode::OK, axum::Json(header_with_proofs)).into_response())
+    } else {
+        Ok(StatusCode::NO_CONTENT.into_response())
+    }
```
merklefruit: I would refactor this to:

```rust
if let Some(best_relay_bid) = state.add_bids(params.slot, relay_bids) {
	let header_with_proofs = GetHeaderWithProofsResponse {
		// ....
	};

	Ok((StatusCode::OK, ...))
} else {
	Ok(StatusCode::NO_CONTENT...)
}
```

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,582 @@
...
+        }
+    }
+
+    let header = state.add_bids(params.slot, relay_bids);
+
+    let header_with_proofs = header.map(|h| GetHeaderWithProofsResponse {
+        data: SignedExecutionPayloadHeaderWithProofs {
+            proofs: hash_to_proofs
+                .get(&h.data.message.header.block_hash)
+                .expect("Saved proofs")
```
merklefruit: what guarantees do we have that `hash_to_proofs` contains as key the block hash of the winning bid? 

Do we assume that all bids ever sent for this slot are the ones in `relay_bids`?
That's fine but it's not super clear in my mind, can you confirm?

thedevbirb (commented):
First round of review! Great work, going to take a better look at the `server.rs`, `types.rs` files later

bolt-boost/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    for relay in &pbs_config.relays {
+        tracing::info!("ID: {} - URI: {}", relay.id, relay.config.entry.url);
+    }
+
+    let custom_state = BuilderState::from_config(extra);
+    let state = PbsState::new(pbs_config).with_data(custom_state);
+
+    // TODO: metrics as below
+    // PbsService::register_metric(Box::new(CHECK_RECEIVED_COUNTER.clone()));
```
merklefruit: there are a bunch of metrics TODOs, should we track them?

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,582 @@
...
+///
+/// The added endpoints are defined in [extra_routes](ConstraintsApi::extra_routes).
+pub struct ConstraintsApi;
+
+#[async_trait]
+impl BuilderApi<BuilderState> for ConstraintsApi {
+    /// Register a validator with the builder.
+    ///
+    /// We intercept this call since it happens periodically and we use it to clean
+    /// up old constraints.
```
merklefruit, line 86: I get why this happens, but what happens if the validator doesn't call this endpoint regularly? we risk going OOM?

merklefruit (changes requested):
Great work, left mostly questions to understand the entire flow

bolt-boost/src/constraints.rs
merklefruit: let's use parking lot!

bolt-boost/src/constraints.rs
mempirate: Thanks for the reminder, done

bolt-boost/src/proofs.rs
mempirate: Added a note! Multiproof generation is not supported in main yet.

bolt-boost/src/proofs.rs
mempirate: Added a note!

bolt-boost/src/server.rs
mempirate: Agreed, this is in commit-boost however

bolt-boost/src/server.rs
mempirate: Fixed, much better

bolt-boost/src/server.rs
mempirate: Added comments. But yes it's possible that there were no proofs if there were no constraints, so fixed the panic!

bolt-boost/src/server.rs
mempirate: Every beacon node registers their validators every epoch, so seems like a pretty big edge case

merklefruit: Great, works for me then!

merklefruit:
One last nit: we're using standardized rustfmt rules across our packages, can you apply them to this one too?
it's just a new file `rustfmt.toml`:

```toml
reorder_imports = true
imports_granularity = "Crate"
use_small_heuristics = "Max"
comment_width = 100
wrap_comments = true
binop_separator = "Back"
trailing_comma = "Vertical"
trailing_semicolon = false
use_field_init_shorthand = true
format_code_in_doc_comments = true
doc_comment_code_block_width = 100
```

then we can use auto-formatting with LSPs and `cargo fmt --all` to do it manually.

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,588 @@
...
+    // Save constraints for the slot to verify proofs against later.
+    for signed_constraints in &constraints {
+        if let Err(e) = state.data.constraints.insert(
+            signed_constraints.message.slot,
+            signed_constraints.message.clone(),
+        ) {
+            error!(slot = signed_constraints.message.slot, error = %e, "Failed to save constraints");
+            return Err(PbsClientError::BadRequest);
+        }
+    }
```
thedevbirb: Related to the comment above on clearing the cache upon registering validators, I think this implementation may introduce some OOM risks:
- the `slot` is not validated, that is we don't check whether as a proposer I'm actually going to propose a block at that slot
- in the underlying hashmap of the cache, every `Vec` value can grow indefinitely.
As such I'm not sure how this endpoint is resistant to spamming. I'd propose an LRU cache or FIFOMap with a certain capacity, and bound checks on vectors not growing past `MAX_CONSTRAINTS_PER_SLOT`

bolt-boost/src/server.rs
```diff
@@ -0,0 +1,588 @@
...
+
+                // If we have constraints to verify, do that here in order to validate the bid
+                if let Some(ref constraints) = maybe_constraints {
+                    // Verify the multiproofs and continue if not valid
+                    if let Err(e) = verify_multiproofs(constraints, &res.data.proofs, root) {
+                        error!(?e, relay_id, "Failed to verify multiproof, skipping bid");
+                        continue;
+                    }
+
+                    tracing::debug!("Verified multiproof in {:?}", start.elapsed());
```
thedevbirb: In a slightly more optimistic setting, isn't it better to verify multiproofs only on the winning bid?

thedevbirb (commented):
A couple of q's on the server code

bolt-sidecar/Cargo.toml
namn-grg: You left this @mempirate  :)

namn-grg: Also wanted to ask why do we need `commit-boost` and `cb-common` separate? cant it be done from single one only?

bolt-boost/src/server.rs
mempirate: Proof verification is not necessarily a bottleneck (takes like 50 microseconds max). It would make it more complicated to go back and find the second best bid in case the best bid is invalid

bolt-boost/src/server.rs
mempirate: Added some simple safeguards to minimize OOM risk

bolt-sidecar/Cargo.toml
mempirate: Not everything is exported through the main module

thedevbirb (approved):

merklefruit (approved):

PR #212, opened 2024-09-05
feat(sidecar): constraints client

### LINKED ISSUE #208: Update constraints-API client
## Context
The sidecar currently supports the old version of the constraints-API. We need to modify the `mevboost` client to support the implementation in
- #207 
- https://chainbound.github.io/bolt-docs/api/builder

Can also rename `mevboost` client to `constraints` client.

namn-grg:
# Context

This PR update the mevboost client to constraint client with support of the new updated constraints api
Closes #208


bolt-sidecar/src/client/constraint_client.rs
```diff
@@ -14,22 +14,26 @@ use crate::{
...
+    primitives::{
+        BatchedSignedConstraints, GetPayloadResponse, SignedBuilderBid, SignedDelegation,
+        SignedRevocation,
+    },
 };
 
 /// A client for interacting with the MEV-Boost API.
 #[derive(Debug, Clone)]
-pub struct MevBoostClient {
+pub struct ConstraintClient {
```
mempirate: Nit: can we rename this to `ConstraintsClient` (and module to `constraints_client`)

bolt-sidecar/src/crypto/bls.rs
```diff
@@ -20,6 +20,9 @@ pub trait SignableBLS {
     /// This API doesn't enforce a specific hash or encoding method.
     fn digest(&self) -> [u8; 32];
```
mempirate, line 20: We can remove the `digest`. The hash tree root should ALWAYS be used to sign.

bolt-sidecar/src/driver.rs
```diff
@@ -105,7 +105,7 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
         commitment_signer: ECDSA,
         fetcher: C,
     ) -> eyre::Result<Self> {
-        let mevboost_client = MevBoostClient::new(cfg.mevboost_url.clone());
+        let mevboost_client = ConstraintClient::new(cfg.mevboost_url.clone());
```
mempirate: Change `mevboost_url` to `constraints_url`

mempirate (changes requested):
Great work! Requested some small changes

bolt-sidecar/src/driver.rs
namn-grg: I felt `constraints_url` will be a bit confusing (cc @merklefruit wdyt? )

bolt-sidecar/src/primitives/constraint.rs
```diff
@@ -58,7 +63,12 @@ impl ConstraintsMessage {
         let constraints =
             request.txs.into_iter().map(|tx| Constraint::from_transaction(tx, None)).collect();
 
-        Self { validator_index, slot: request.slot, constraints }
+        Self { validator_index, slot: request.slot, top: false, constraints }
+    }
+
+    /// Returns the total number of leaves in the tree.
+    fn total_leaves(&self) -> usize {
+        4 + self.constraints.len()
```
merklefruit: why do we start at 4?

bolt-sidecar/src/driver.rs
merklefruit: `constraints_url` is more idiomatic considering mev-boost is "implementation" rather than spec.
Just like in the CL you have to specify `--builder` to point to mev-boost because it refers to the builder API.

merklefruit: ref https://lighthouse-book.sigmaprime.io/builders.html#how-to-connect-to-a-builder

bolt-sidecar/src/primitives/constraint.rs
namn-grg: Not sure, I took it from the bolt-boost impl by @mempirate

thedevbirb: My bad I haven't noticed it in his PR, but I'm pretty sure it is because the way you encode the struct (which has $$4$$ properties, along with a list) in a binary merkle tree.
@mempirate could you please confirm and document the number in the `bolt-boost` impl as well?

thedevbirb: Also I'm not sure about this computation since it should take into account chunking and padding: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#merkleization

mempirate: We have 3 "unit" types + a list, which will add another leaf for the length + leaves for each individual elements! Will document

mempirate: I realized I haven't tested this - @namn-grg could you add a quick test that generates a random `ConstraintsMessage` and calculates the hash tree root?

namn-grg: @mempirate added lmk how it looks?

merklefruit: LGTM, we can improve it with the use of `arbitrary` and `arbtest` for prop testing but random is good enough for now.

thedevbirb: This test only asserts that `hash_root` is 32 bytes long but that's trivial due to the signature of the `hash_tree_root` method. @mempirate how do we intend to test this? 
In my opinion we should compare the output of this implementation with what fastssz outputs once having generated the hash tree root implementation for this struct using codegen: https://github.com/ferranbt/fastssz/blob/main/sszgen/main.go

bolt-sidecar/src/primitives/constraint.rs
```diff
@@ -41,72 +43,102 @@ pub struct SignedConstraints {
...
     /// The validator index of the proposer sidecar.
     pub validator_index: u64,
     /// The consensus slot at which the constraints are valid
     pub slot: u64,
+    /// Indicates whether these constraints are only valid on the top of the block.
+    /// NOTE: Per slot, only 1 top-of-block bundle is valid.
+    pub top: bool,
     /// The constraints that need to be signed.
-    pub constraints: Vec<Constraint>,
+    pub constraints: Vec<FullTransaction>,
```
mempirate, line 62: We should have a custom serialize / deserialize function here that uses EIP-2718 encoding. Let's also add a rountrip ser / de test.

namn-grg: Thanks for poiniting this out, done!

bolt-sidecar/src/primitives/mod.rs
```diff
@@ -381,22 +381,34 @@ impl FullTransaction {
...
 }
 
-impl<'de> serde::Deserialize<'de> for FullTransaction {
-    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        let s = String::deserialize(deserializer)?;
+fn deserialize_txs<'de, D>(deserializer: D) -> Result<Vec<FullTransaction>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let hex_strings = <Vec<String> as de::Deserialize>::deserialize(deserializer)?;
```
mempirate: Replace `String` with `&str` here to avoid extra allocation

bolt-sidecar/src/primitives/mod.rs
```diff
@@ -381,22 +381,34 @@ impl FullTransaction {
...
+    let mut txs = Vec::new();
```
mempirate: Initialize with capacity here: `Vec::with_capacity(hex_strings.len())`

mempirate (changes requested):
Some final perf improvements requested

bolt-sidecar/src/primitives/mod.rs
```diff
@@ -376,29 +376,65 @@ impl FullTransaction {
     }
 
     /// Returns the sender of the transaction, if recovered.
-    pub fn sender(&self) -> Option<Address> {
-        self.sender
+    pub fn sender(&self) -> Address {
+        self.sender.expect("Recovered sender")
```
merklefruit: I'm not a huge fan of quiet panics, let's think of an improvement to this in a later PR.
Not blocking.

mempirate: Reverted

PR #223, opened 2024-09-11
feat(contracts): new challenger

### LINKED ISSUE #199: Challenger implementation V1
## Context
For Holesky we want an initial working challenger implementation (without actual slashing).

merklefruit:
### Overview

This iteration of the `BoltChallenger` implements the new Commitment Challenge game.
The challenge resolution flow consists of the following steps:

1. identify the challengeID you want to solve, based on the event emitted upon new challenge creation
2. provide a `blockHash` from a trusted source. For recent blocks, this can be the EVM directly. For historical blocks, this can come from a number of sources like blockhash accumulators, zk-coprocessors, or other proving systems.
3. provide a proof consisting of:
    * RLP-encoded block header (hashing to the provided blockHash)
    * Account merkle trie proof
    * Transaction merkle trie proof

### Tasks

- [x] decode block header, given a trusted block hash
- [x] prove existence of an account and decode its contents from RLP, given a proof and a trusted state root
- [x] prove inclusion of a transaction and decode its contents from RLP, given a proof and a trusted tx root
- [x] add unit tests for opening challenges
- [x] build an end2end test with the challenge resolution proof flow
- [x] build out the functionality to open a new challenge
- [x] build out the first challenge resolution logic
- [x] add more unit tests for fault resolution logic
- [x] add extensive documentation

### RLP transaction decoder

- [x] implement decoding for EIP-2718 transaction envelopes of type 0, 1, 2, 3
- [x] test decoding & preimages with ethers-js test vectors

Extra (out of scope):

- write a Rust library and CLI tool to generate all the required proofs for resolving a challenge
- write a challenger agent binary that monitors the chain for newly opened challenges

### Meta

- closes #199


bolt-contracts/README.md
thedevbirb, line 1: This is great, I'd just highlight what are the currently supported challenge times:
- after a preconf has been requested for a slot we need to wait some time before the block has been finalized
- after 256 slots it not possible to request a challenge

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+            abi.encodePacked(decodedTx.recoverSender()), proof.accountMerkleProof, blockHeader.stateRoot
+        );
+
+        if (!accountExists) {
+            revert AccountDoesNotExist();
+        }
+
+        AccountData memory account = _decodeAccountRLP(accountRLP);
+
+        if (account.nonce > decodedTx.nonce) {
```
thedevbirb: I'm not sure about this check:
- Suppose my account before sending the preconf has nonce $n$. As such, the nonce of the tx is $n$ as well. After it has been included my account will have nonce $n+1$. Since the `proof.accountMerkleProof` is performed against the `blockHeader.stateRoot` after all its transaction I'm worried we might always enter this `if` condition is always satisfied
- What if I send multiple preconfirmations for the same slot? According to the reasoning of the previous point, can we encounter some edge cases?

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+            // The transaction recovered sender has sent a transaction with a higher nonce than the committed
+            // transaction, before the proposer could include it. Consider the challenge defended, as the
+            // proposer is not at fault. The bond will be shared between the resolver and commitment signer.
+            challenge.status = ChallengeStatus.Defended;
+            _transferHalfBond(msg.sender);
+            _transferHalfBond(challenge.target);
+            emit ChallengeDefended(challengeID);
```
thedevbirb: Transferring part of the bond to an entity different than the commitment signer (which is the user who requested the preconf) might incentive frontrunning the user, doesn't it?
I think the user has already all the incentives to do it by itself.

This is kinda an open question so I'd drop a TODO comment above this logic

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+            return;
+        } else if (account.nonce < decodedTx.nonce) {
+            // Q: is this a valid case? technically the proposer would be at fault for accepting a commitment of an
+            // already included transaction. TBD.
+        }
```
thedevbirb: If such condition is `true` then it means the proposer has committed to an invalid transaction (when it was requested) and didn't return a `NonceTooHigh` error. This is probably an internal error of our sidecar and as such I don't think the proposer should be slashed.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+
+        if (account.balance < blockHeader.baseFee * decodedTx.gasLimit) {
+            // The account does not have enough balance to pay for the worst-case base fee of the committed transaction.
+            // Consider the challenge defended, as the proposer is not at fault. The bond will be shared between the
+            // resolver and commitment signer.
+            challenge.status = ChallengeStatus.Defended;
+            _transferHalfBond(msg.sender);
+            _transferHalfBond(challenge.target);
```
thedevbirb: Same reasoning of the previous comment

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+            emit ChallengeDefended(challengeID);
+            return;
+        } else if (account.nonce < decodedTx.nonce) {
+            // Q: is this a valid case? technically the proposer would be at fault for accepting a commitment of an
+            // already included transaction. TBD.
+        }
+
+        if (account.balance < blockHeader.baseFee * decodedTx.gasLimit) {
```
thedevbirb: Kinda related to the nonce check: if the account state considered _after all the transactions have been applied_? If so, this check might return true because the user might have consumed all its balance even if it has actually paid for the preconf. 
If the account state is _before transactions have been applied_, could you please document it very clearly? This has a great impact on the dispute logic.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+            emit ChallengeDefended(challengeID);
+            return;
+        }
+
+        // The key in the transaction trie is the RLP-encoded index of the transaction in the block
+        bytes memory txLeaf = RLPWriter.writeUint(proof.txIndexInBlock);
+
+        // Verify transaction inclusion proof
+        // Note: the transactions trie is built with raw leaves, without hashing them first.
+        // This denotes why we use `MerkleTrie.get()` as opposed to `SecureMerkleTrie.get()` here.
```
thedevbirb: That was a fun afternoon of debugging :)

bolt-contracts/src/interfaces/IBoltChallenger.sol
```diff
@@ -0,0 +1,81 @@
...
+        Defended,
+        Lost
+    }
+
+    struct Challenge {
+        bytes32 id;
+        uint48 openedAt;
+        ChallengeStatus status;
+        address challenger;
+        address target;
```
thedevbirb: Perhaps some docs on these fields could be helpful -- I haven't very clear who is the `target` for example

bolt-contracts/src/lib/TransactionDecoder.sol
```diff
@@ -0,0 +1,451 @@
...
+    using RLPReader for bytes;
+    using RLPReader for RLPReader.RLPItem;
+
+    /// @notice The type of transaction
+    enum TxType {
+        Legacy,
+        Eip2930,
+        Eip1559,
+        Eip4844
+    }
```
thedevbirb, line 24: How do we handle hard-fork updates with this library code?

bolt-contracts/src/lib/TransactionDecoder.sol
thedevbirb, line 1: Amazing job here, as outlined in the comment on the `enum` I'm generally curious how can we handle hard-fork updates for this code.

thedevbirb (commented):
Stunning work. Just left some comments on the dispute logic since I've trouble understanding to what data the account proofs points to: _before transactions_ or _after transactions_?

bolt-contracts/README.md
merklefruit: Good point, will add rn. I'll also mention that after EIP-2935 we can extend that to ~1 day #237

merklefruit: I've also changed the finalization_delay to a justification_delay, as in accordance with Ethereum consensus it's not possible to reorg a previously justified checkpoint. A 32-deep reorg would be _possible_ without breaking Ethereum, but it would be a huge black swan event and breaking a Bolt challenge doesn't seem worth all that effort :)

bolt-contracts/src/interfaces/IBoltChallenger.sol
merklefruit: thanks - changed to `commitmentSigner` - this is the party that signed the user's commitment aka the operator

bolt-contracts/src/lib/TransactionDecoder.sol
merklefruit: that's the fun part: we don't :)

The best IMO would be to simply keep track of the challenger implementation in BoltManager and make it upgradeable, so that we can upgrade it if:

a. we find logic bugs / exploits
b. we need to add new types (such as for new hardforks)
c. we want to add a new challenge type or functionality (e.g. zk-coprocessor resolution? or anything else)

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: > I think the user has already all the incentives to do it by itself.

it does incentivise front-running, hence the question comment on the _resolve 
function if we should use a commit and reveal scheme. 

This incentive makes sense in my mind because we want a healthy and speedy processing of challenges.
From a mechanism perspective it's better to give a reward to the resolver (even if less than 0.5 ETH) IMO

thedevbirb: Ok I get this can get out of scope but I have some considerations. A commit and reveal scheme could be cool but it's added complexity we could avoid if possible. I'd personally strive for simplicity in the code with one of the two approaches:
- Just give everything to the user who requested the preconf. I think there are easy ways to handle this while retaining sufficiently nice UX. 
- if `msg.sender != challenge.target` or the user, we could give back a fixed percentage over `base_fee * gas_limit`, where `gas_limit` must be lower a certain bound we know after some simulation. This incentivise speedy processing but by giving only marginal gains, akin to having solvers.

thedevbirb: I'll solve the comment for now and we can discuss this further privately

bolt-contracts/src/lib/TransactionDecoder.sol
thedevbirb: I like this, can we track it in an issue or somewhere else?

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: You're right, the current challenge doesn't handle stacked preconfs in the same slot.
This will need more attention, moving convo offline

merklefruit: Addressed in #239

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -0,0 +1,357 @@
...
+
+        // Finally resolve the challenge with the trusted block hash and the provided proofs
+        _resolve(challengeID, trustedBlockHash, proof);
+    }
+
+    // Resolving a historical challenge requires acquiring a block hash from an alternative source
+    // from the EVM. This is because the BLOCKHASH opcode is limited to the 256 most recent blocks.
+    function resolveChallenge(bytes32 challengeID, Proof calldata proof) public {
+        // unimplemented!();
+    }
```
mempirate: Is this still to do?

bolt-contracts/src/contracts/BoltChallenger.sol
mempirate: Good points! A simple fix is using the previous blocks' state root to do these checks right?

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: no it's outdated

PR #202, opened 2024-08-23
feat(contracts): new registry (V1)

merklefruit:
## Commitment registry with external collateral

Closes #196 
Closes #36 
Closes #244 
Closes #273 

### Registration procedure

1. Validator registers in `BoltValidators`. There is a `registerValidatorUnsafe` method that skips all verifications (BLS signature check and SSZ validator existence check). This function will be useful to test and until Pectra will be live on public testnets. We can also start testing with the precompile on Pectra EF devnets for now.
2. Operators register in `BoltManager` via `registerSymbioticOperator()`
3. Operators do all of their Symbiotic-related authorization steps on the `Vaults` they want to operate on
4. Vaults register in `BoltManager` via `registerSymbioticVault()` (currently no check on collateral asset used)

The `BoltManager` contract allows read access into the current stake of any operator in the network
at any given time, for any asset:

```solidity
function getSymbioticOperatorStakeAt(
		address operator,   // the operator for which to get the stake
        address collateral, // the collateral asset address to query for
        uint48 timestamp    // the timestamp at which the query is made
    ) external view returns (uint256);
```

It also allows to query if an operator is authorized to emit commitments for a specific validator:

```solidity
function isSymbioticOperatorAuthorizedForValidator(
        address operator,
        bytes32 pubkeyHash
    ) external view returns (bool);
```

The actual staking, unstaking, and slashing downstream logics are left to Symbiotic for now.

### Diagrams

shapes at 24-09-03 10 51 15

### Missing & TODO

- [x] should we gate the collateral assets behind a whitelist? 
- [x] ability to verify aggregated bls12_381 signature for batch registration
- [x] ability for `Validator`s to pause or change the currently authorized `Operator`

### Meta

- reference discussion: #95 
- part of #196 

## Summary by CodeRabbit

- **New Features**
	- Introduced a Continuous Integration (CI) workflow for automated testing of smart contracts.
	- Added new mock contracts for testing collateralized debt positions and ERC20 tokens.
	- Enhanced documentation with a structured Table of Contents and detailed integration guides.

- **Bug Fixes**
	- Improved error handling for non-existent operators and validators in the testing framework.

- **Documentation**
	- Expanded README with comprehensive explanations of the Bolt ecosystem components and security considerations.

- **Tests**
	- Implemented a robust testing suite for `BoltManager` and `BoltValidators`, validating key functionalities and interactions.

PR #224, opened 2024-09-11
Commitment Registry: EigenLayer support (take 2)

thedevbirb:
## What
This PR introduces some functions inside the `BoltManager` contract to allow integration via EigenLayer:
- `registerEigenLayerOperatorToAVS`,
- `checkIfEigenLayerOperatorRegisteredToAVS`,
- `deregisterEigenLayerOperatorFromAVS`,
- `getEigenLayerOperatorStake`,

Along with all the other functions needed for interacting with the `BoltManager` and `BoltValidators` contracts, which are essentially a duplicate of the logic already written for Symbiotic. See https://github.com/chainbound/bolt/pull/202.

## Flow
Such functions have been introduced according to EigenLayer architecture and registration flow. Since it requires some effort to get familiar with re-staking protocols, we recap it here along with some useful resources/references.

**Useful resources**
- Key terms and Definitions: https://docs.eigenlayer.xyz/eigenlayer/overview/key-terms
- AVS book: https://eigenlabs.gitbook.io/avs-book
- Awesome AVS: https://github.com/Layr-Labs/awesome-avs

**Vanilla EL flow step by step**

1. As a staker, I deposit some LSTs into a Stategy (a SC abstraction over an asset re-staked into EigenLayer) via the [`StrategyManager.depositIntoStrategy`](https://github.com/Layr-Labs/eigenlayer-contracts/blob/mainnet/src/contracts/core/StrategyManager.sol#L104-L110) function. After this, I get back some *shares* that I can use at a later time for withdrawal
2. As a Operator, I register myself into EigenLayer using [`DelegationManager.registerAsOperator`](https://github.com/Layr-Labs/eigenlayer-contracts/blob/mainnet/src/contracts/core/DelegationManager.sol#L107-L119). Note that this function doesn’t require specifying anything related to the service I’m going to provide. However, a parameter describes who can delegate to me whether it can be anyone or a subset of stakers.
3. As a staker, I can start delegating funds to these operators using the [`DelegationManager.delegateTo`](https://github.com/Layr-Labs/eigenlayer-contracts/blob/mainnet/src/contracts/core/DelegationManager.sol#L154-L163) function and specifying to who I wish to delegate my funds. Multiple delegator support is not available at the moment
4. As an AVS developer I create an entrypoint contract, usually called `${ServiceName}Manager` (from now on, just `ServiceManager`) . Upon deploying the contract it is required to make a call to EL’s [`AVSDirectory.updateAVSMetadataURI`](https://github.com/Layr-Labs/eigenlayer-contracts/blob/mainnet/src/contracts/core/AVSDirectory.sol#L133-L135) which takes just a string which is a URI. Note that his is not stored anywhere, just an log is emitted. Note that `msg.sender` which is the `ServiceManager` contract is used to identify the AVS itself
5. As a operator, I can now opt-in into an AVS by interacting with the `ServiceManager` . Two steps happen:
    1. I call the AVS’ `ServiceManager.registerOperatorToAVS` . The payload is a signature whose digest consists of:
        1. my operator address
        2. the AVS’ `ServiceManager` contract address
        3. a salt
        4. an expiry
    2. The contract forwards the call to the [`AVSDirectory.registerOperatorToAVS`](https://github.com/Layr-Labs/eigenlayer-contracts/blob/mainnet/src/contracts/core/AVSDirectory.sol#L64-L108) to that `msg.sender` is the AVS contract. Upon successful verification of the signature, the operator is considered `REGISTERED`  in a mapping `avsOperatorStatus[msg.sender][operator]`.
6. As a user, I can issue a computation/service by leveraging the implementation of `ServiceManager`/ the AVS protocol.

In our implementation `ServiceManager` is `BoltManager`, and `registerOperatorToAVS` is `registerEigenLayerOperatorToAVS`.

**Observations**:

- Shares do not accrue value, they’re just a mean to represent proportional ownership of the assets.
- Due to the previous point, EigenLayer is agnostic of the payments to operators.

**Diagrams**
For consistency, we report an analogous version of the Bolt Symbiotic opt-in flow but for EigenLayer:
image
Note that we assume the "Vanilla EL flow" is respected. As such, the operator will have the collateral delegated by the staker/validator which lives on a certain strategy. 

## TODO

I could add a little more tests, however I'd like to do that after this PR for two reasons:
- I'd avoid a big pending PR for possible rebasing issues
- As you can see by the changes of the interface, there is a good need for a middleware abstraction. As such I think it'd be best to do in-depth tests after it to avoid another round of refactoring.


merklefruit (approved):
Looks great. From a first review everything seems fine, but more thorough tests on the flow will be necessary. Closing this PR to reduce rebasing overhead works for me.

Issue #230, opened 2024-09-13
Update constraints-API to use pubkey instead of index


mempirate:
## Context
In the constraints-API we've used the `validatorIndex` to link validators to their signature. This works if the relay has a mapping from `validatorIndex => pubkey`, which not all of them do. So we should modify the constraints-API to use `pubkey` instead of `validatorIndex` in the following endpoints:
- `submit_constraints`
- `delegate`
- `revoke`

We should also update the API docs to reflect this: https://chainbound.github.io/bolt-docs/api/builder

```[tasklist]
# Tasks
- [x] Update `submit_constraints`
- [x] Update `delegate`
- [x] Update `revoke`
- [ ] Update types across PBS stack
- [x] Update documentation
```

namn-grg:
Changes in the builder can be done together in #209 when updating builder for the new constraint api (cc @merklefruit )

PR #234, opened 2024-09-17
chore(sidecar): use validator pubkey instead of index

namn-grg:
Solves #230 

This PR updates the sidecar to use validator pubkey instead of the index


merklefruit (approved):
LGTM

thedevbirb (approved):

PR #239, opened 2024-09-25
feat(challenger): added stacked preconfs in the same slot

merklefruit:
### Overview

This PR adds the logic to handle an important missing piece of the challenger logic: stacked preconfirmations.

A couple useful definitions:

- `commitmentReceiver`: this is the EOA that requested the preconfirmation. it is the sender of the signed transaction that has been preconfirmed by a proposer.
- `commitmentSigner`: the party that signed the commitment. This is anyone running a Bolt sidecar, aka proposers.

### The problem

The issue, in short, can be summarized by the following example:

- Alice requests 3 preconfirmations for the same slot from the same proposer
- The proposer accepts all of them and signs them, but then includes only 1 and 2, skipping 3.
- Alice would try to open a challenge, only to see that their nonce was TOO_HIGH, because the challenger contract has no way to prove that transactions 1 and 2 were preconfirmed, as the challenge was for a single committed transaction.

### The solution

To address this, we now require the challenge to be opened with an array of committed transactions.
These transactions cannot be random though, they must follow these criteria:

- they must be sent by the same entity: a "stack" is valid only for each individual EOA, as we really only care about the sender's nonce and balance in order to validate the protocol rules for inclusion
- they must be signed by the same bolt proposer, which we recover through EC recovery
- they must target the same slot, which is part of all signed commitments
- they must have strictly sequential (monotonically increasing) nonces

Then, when it comes to verifying their inclusion, we check each transaction's nonce one after the other to validate them, before performing the usual merkle inclusion proof. 

Currently the challenge is considered BREACHED if any of the commitments inside the bundle is breached. 
A further improvement could keep track of the number of breached commitments easily, but that's out of scope here.

### How it works

The main change this PR introduces is a new structure for the `Challenge` and `Proofs` structs:

```solidity
    struct Challenge {
        bytes32 id;
        uint48 openedAt;
        ChallengeStatus status;
        uint256 targetSlot; // new
        address challenger;
        address commitmentSigner;
        address commitmentReceiver; // new
        TransactionData[] committedTxs; // previously this was not an array
    }
```

- `commitmentReceiver` explained above
- `targetSlot` refers to the slot contained in all the commitments of this challenge
- `committedTxs` contains data about each individual committed transaction, like its nonce and gas limit.

```solidity
    struct Proof {
        uint256 inclusionBlockNumber; // same as before, prefixed with "inclusion" for clarity
        bytes previousBlockHeaderRLP; // block against which we check initial nonce and balance
        bytes inclusionBlockHeaderRLP; // block against which we check committed transaction inclusion
        bytes accountMerkleProof;
        bytes[] txMerkleProofs; // this is now an array, must have length equal to Challenge.committedTxs
        uint256[] txIndexesInBlock; // same as above
    }
```

### Meta

- [x] updated challenger interface and implementation
- [x] updated existing tests
- [ ] updated documentation


bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
-        SignedCommitment calldata commitment
+        SignedCommitment[] calldata commitments
     ) public payable {
+        if (commitments.length == 0) {
+            revert EmptyCommitments();
+        }
+
         // Check that the challenge bond is sufficient
-        if (msg.value < CHALLENGE_BOND) {
+        uint256 expectedBond = COMMITMENT_BOND * commitments.length;
```
thedevbirb: Can you motivate this choice further? I don't think this has been discussed internally and it can lead to an high bond even with a few preconfirmations asked for a slot.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
+        if (msg.value < expectedBond) {
             revert InsufficientChallengeBond();
-        } else if (msg.value > CHALLENGE_BOND) {
+        } else if (msg.value > expectedBond) {
             // Refund the excess value, if any
-            payable(msg.sender).transfer(msg.value - CHALLENGE_BOND);
+            payable(msg.sender).transfer(msg.value - expectedBond);
         }
```
thedevbirb: This is a nit but what's the advantage of giving back excess value instead of reverting when `msg.value != expectedBond`? The users of this function are going to simulate it against their inputs nonetheless so it's better to give them a feedback earlier instead of making them pay a bit more than 21000 gas at runtime.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
         });
-
-        emit ChallengeOpened(commitmentID, msg.sender, commitmentSigner);
+        emit ChallengeOpened(challengeID, msg.sender, commitmentSigner);
     }
 
     // ========= CHALLENGE RESOLUTION =========
 
-    function resolveRecentChallenge(bytes32 challengeID, Proof calldata proof) public {
+    function resolveChallenge(bytes32 challengeID, Proof calldata proof) public {
```
thedevbirb: Can you add some simple documentation for this? Perhaps with a reference to the `Proof` data and what it contains. I see you did it for the internal `_resolve` function but it might be helpful here as well.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
             revert BlockIsTooOld();
         }
 
-        // Check that the block number is within the EVM lookback window for block hashes
-        if (proof.blockNumber > block.number || proof.blockNumber < block.number - BLOCKHASH_EVM_LOOKBACK) {
+        // Check that the previous block is within the EVM lookback window for block hashes.
+        if (
+            (proof.inclusionBlockNumber - 1) > block.number
+                || (proof.inclusionBlockNumber - 1) < block.number - BLOCKHASH_EVM_LOOKBACK
+        ) {
```
thedevbirb: A nit just to make it easier to read and save some smol gas.
```suggestion
        // Check that the previous block is within the EVM lookback window for block hashes.
        uint256 previousInclusionBlockNumber = proof.inclusionBlockNumber - 1;
        if (
            previousInclusionBlockNumber > block.number
                || previousInclusionBlockNumber < block.number - BLOCKHASH_EVM_LOOKBACK
        ) {
```

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -194,101 +232,138 @@ contract BoltChallenger is IBoltChallenger {
...
+
+        // Decode the RLP-encoded block header of the inclusion block.
+        //
+        // The inclusion block is necessary to extract the transaction root and verify the inclusion of the
+        // committed transactions. By checking against the previous block's parent hash we can ensure this
+        // is the correct block trusting a single block hash.
+        BlockHeaderData memory inclusionBlockHeader = _decodeBlockHeaderRLP(proof.inclusionBlockHeaderRLP);
+
+        // Check that the inclusion block is a child of the previous block
+        if (inclusionBlockHeader.parentHash != keccak256(proof.previousBlockHeaderRLP)) {
```
thedevbirb: Can you store `keccak256(proof.previousBlockHeaderRLP)` in a variable? Since this is also re-used above

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -31,7 +31,7 @@ contract BoltChallenger is IBoltChallenger {
     // ========= CONSTANTS =========
 
     /// @notice The challenge bond required to open a challenge.
-    uint256 public constant CHALLENGE_BOND = 1 ether;
+    uint256 public constant COMMITMENT_BOND = 1 ether;
```
mempirate: Why was this renamed? I think `CHALLENGE_BOND` is more clear since it's the bond required to open a challenge

bolt-contracts/test/BoltChallenger.t.sol
thedevbirb, line 1: Can you add a test also for stacked preconfirmations?

thedevbirb (commented):
The logic is very clear and that's great. Left some observations.

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
+        if (challengeIDs.contains(challengeID)) {
             revert ChallengeAlreadyExists();
         }
 
         // Add the challenge to the set of challenges
-        challengeIDs.add(commitmentID);
-        challenges[commitmentID] = Challenge({
-            id: commitmentID,
+        challengeIDs.add(challengeID);
+        challenges[challengeID] = Challenge({
```
mempirate, line 168: Do we clear expired challenges somewhere?

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: Good call, yeah it doesn't make a difference in practice, I can change this

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: No, they live in the storage forever, as we rely on this as an oracle for understanding if a slashing request is legit.

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: After thinking about it you're right, it's just anti-spam so it doesn't need to scale with commitments. 
I was just thinking that since resolving bigger challenges requires more gas you'd want a bigger refund to defend.

So far I haven't made a test of the incremental gas cost to include 2+ signed commitments in the resolution tho.

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: see above comment by Lore: https://github.com/chainbound/bolt/pull/239#discussion_r1775134433

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: Ok will do, however please don't judge the gas efficiency of the implementation too much overall, I know there's a lot that can be cut but I wanted to focus on the core logic

bolt-contracts/src/contracts/BoltChallenger.sol
thedevbirb: Just wondering: after a challenge has been resolved can we keep its ID in the `challengeIDs` storage but remove it from `challenges`? We don't need that data anymore and can also give some gas refunds if I understood correctly (https://www.evm.codes/?fork=cancun#55).

With this, then we can perform this check on `_resolve`:
```solidity
    function _resolve(bytes32 challengeID, bytes32 trustedPreviousBlockHash, Proof calldata proof) internal {
        if (!challengeIDs.contains(challengeID)) {
            revert ChallengeDoesNotExist();
        }

		// This means the challenge has been resolved and deleted from storage successfully.
        if ( challenges[challengeID].id == bytes32(0)) {
            revert ChallengeAlreadyResolved();
        }

		// ...
```

merklefruit: Let me make a test to see how much gas this would save

merklefruit: Before: `BoltChallengerTest:testResolveChallengeFullDefenseSingleTx() (gas: 697302)`
After: `BoltChallengerTest:testResolveChallengeFullDefenseSingleTx() (gas: 554812)`

Savings: 142.490 gas which at ~30 gwei would cost like $10 at current ETH prices.
This will be even more for N constraints. I think overall it makes sense and I'll think of a better solution
to keep track of the basic necessary data and get rid of everything else!

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -98,92 +98,130 @@ contract BoltChallenger is IBoltChallenger {
...
+            // check that the nonces are strictly sequentially increasing in the bundle
+            if (otherTransactionData.nonce != transactionsData[i - 1].nonce + 1) {
+                revert UnexpectedNonceOrder();
+            }
+        }
 
-        // Verify the commitment signature against the digest
-        address commitmentSigner = ECDSA.recover(commitmentID, commitment.signature);
+        // Build the challenge ID: `keccak( keccak(signed tx 1) || keccak(signed tx 2) || ... || le_bytes(slot) )`
+        bytes32 challengeID = _computeChallengeID(commitments);
```
merklefruit: I'm not 100% sure this works well - TBD

bolt-contracts/test/BoltChallenger.t.sol
merklefruit: done!

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: this leaves an edge case open:
* proposer commits to 10 transactions (same sender, same slot)
* proposer includes only the first 3, the last 7 are broken commitments
* problem: challengeID should be unique for that sender and slot, but the challenge could be opened with any amount of signed commitments as calldata
* we can't guarantee that the first time the challenge is opened it will contain the full list of broken commitments
* it would be currently possible to open the challenge with commitments (1, 2, 3, 4), or (1, 2, 3, 4, 5), etc until (1, 2, 3, 4, 5, 6, 7, 8, 9, 10) and all of them would be considered lost because there is at least a broken commitment in each.

A granular solution would be to keep track of all the tx hashes in storage in a mapping of (bytes32 -> bool) and slashing will be done at most once per tx hash for which the mapping is true. Thinking of possible alternatives.

thedevbirb: I've seen that now the signature is computing by concatenating signatures which is good but still leaves this edge case exposed. I remember the latest proposal on it is to create a challenge _per_ commitment breached, but I don't see this logic in the contract otherwise the function `_computeChallengeID` would return `bytes32[] memory challengeIDs`. Is there any reason we're not doing it?

thedevbirb (approved):

bolt-contracts/src/contracts/BoltChallenger.sol
merklefruit: As discussed privately, this will be addressed in a later iteration. tracked in #248

PR #240, opened 2024-09-25
chore(sidecar): update constraint api

namn-grg:
This PR -
- renames the `constraints` field to `transactions` 
- adds a test data generator for testing

### DIFF
--- a/bolt-sidecar/src/crypto/bls.rs
+++ b/bolt-sidecar/src/crypto/bls.rs
@@ -102,11 +102,10 @@ fn sign_with_prefix(key: &BlsSecretKey, data: &[u8]) -> Signature {
 #[cfg(test)]
 mod tests {
     use crate::{
-        crypto::bls::{SignableBLS, Signer, SignerBLS}, primitives::{ConstraintsMessage, DelegationMessage, FullTransaction, InclusionRequest, RevocationMessage, SignedConstraints, SignedDelegation, SignedRevocation}, test_util::{test_bls_secret_key, TestSignableData}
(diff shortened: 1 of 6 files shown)


bolt-sidecar/src/crypto/bls.rs
```diff
@@ -124,4 +125,94 @@ mod tests {
...
+            "txs": ["0x02f86c870c72dd9d5e883e4d0183408f2382520894d2e2adf7177b7a8afddbc12d1634cf23ea1a71020180c001a08556dcfea479b34675db3fe08e29486fe719c2b22f6b0c1741ecbbdce4575cc6a01cd48009ccafd6b9f1290bbe2ceea268f94101d1d322c787018423ebcbc87ab4"]
+        }"#;
+
+        let req: InclusionRequest = serde_json::from_str(json_req).unwrap();
+
+        (0..count).map(|_| req.txs.first().unwrap().clone()).collect()
+    }
+
+    #[tokio::test]
+    async fn generate_test_data() {
```
merklefruit: I understand why you created this test and it looks good, however I would suggest moving it out of the `bls` module. Not fully sure where yet, but here it's a bit confusing IMO

namn-grg: yeah I agree. I think `src/test_util.rs` might be a good place? I just thought of keeping this because will be helpful in future

merklefruit: Sure, perfect

namn-grg: done

Issue #209, opened 2024-09-04
Tracking: update constraints-API in PBS


mempirate:
## Context
This issue tracks the implementation of the new constraints API across the PBS pipeline. Note that we can implement Blob Express Lane in the relay & builder.

```[tasklist]
## Bolt Sidecar
- [ ] #208
- [ ] https://github.com/chainbound/bolt/pull/225
- [ ] https://github.com/chainbound/bolt/issues/230
- [ ] https://github.com/chainbound/bolt/issues/290
- [ ] https://github.com/chainbound/bolt/issues/283
```

@merklefruit can work on this:
```[tasklist]
## Flashbots Builder
- [x] Update types
- [x] Remove authentication header (not needed according to the relays)
- [x] Update constraints stream (new format & types)
- [x] Nit: get rid of "preconf" wording everywhere, change to constraints
- [x] Remove Bolt demo events
- [x] Add relevant log lines
```

@namn-grg is working on this:
```[tasklist]
## Helix
- [x] Implement proposer-facing [constraints-API](https://chainbound.github.io/bolt-docs/api/builder)
- [x] Implement builder-facing [constraints-API](https://chainbound.github.io/bolt-docs/api/relay)
- [x] Verify BLS signatures on `SignedConstraints` (will be signed with commit-boost signing domain!)
- [x] Verify authorized BLS pubkey on `SignedConstraints`!
```

PR #241, opened 2024-09-25
Builder: update to new Constraints API

thedevbirb:
This PR does the steps outlined in https://github.com/chainbound/bolt/issues/209 to support the new constraints API

To do: ping Kubi when this is merged


mempirate:
Will not support the `top` field in this PR as we won't support it any time soon any way

builder/builder/builder.go
```diff
@@ -359,14 +362,21 @@ func (b *Builder) subscribeToRelayForConstraints(relayBaseEndpoint string) error
 		}
 
 		for _, constraint := range constraintsSigned {
+			// Check if the signature is valid against any of the authorized pubkeys
+			for _, pubkey := range b.slotConstraintsPubkeys {
+				valid, err := constraint.VerifySignature(pubkey)
+				if err != nil || !valid {
+					log.Error("Failed to verify constraint signature", "err", err)
+					continue
+				}
```
merklefruit, line 372: Why do we just continue in this case?

builder/builder/builder.go
```diff
@@ -570,6 +572,40 @@ func (b *Builder) OnPayloadAttribute(attrs *types.BuilderPayloadAttributes) erro
...
+	constraintsPubkeys := []phase0.BLSPubKey{pubkey}
+
+	// BOLT: get delegations for the slot
+	delegations, err := b.relay.GetDelegationsForSlot(attrs.Slot)
+	if err != nil {
+		log.Error("could not get delegations for slot, using default validator key", "slot", attrs.Slot, "err", err)
+	}
+
+	if len(delegations) > 0 {
+		// If there are delegations, reset the constraintsPubkeys (i.e. remove the validator key as authorized)
```
merklefruit, line 615: So when delegating to a key, the proposer key is revoked? If so, why? Is this outlined in the API specs too?

merklefruit: This doesn't fully make sense to me because the proposer can always issue more delegations making it possible to have more than 1 key authorized to send constraints at a time (which is the reason why we have a list of delegations here). So why would we revoke permissions to the proposer key?

builder/builder/relay.go
```diff
@@ -129,6 +129,28 @@ func (r *RemoteRelay) GetValidatorForSlot(nextSlot uint64) (ValidatorData, error
...
+	endpoint := r.config.Endpoint + fmt.Sprintf("/relay/v1/builder/delegations?slot=%d", nextSlot)
+
+	if r.config.SszEnabled {
+		panic("ssz not supported")
+	}
+
+	var dst common.SignedDelegations
+	code, err := SendHTTPRequest(context.TODO(), *http.DefaultClient, http.MethodGet, endpoint, nil, &dst)
+	if err != nil {
+		return nil, fmt.Errorf("error sending http request block with proofs to relay %s. err: %w", r.config.Endpoint, err)
```
merklefruit: old comment here

builder/common/types.go
```diff
@@ -597,40 +616,88 @@ func (v *VersionedSubmitBlockRequestWithProofs) String() string {
...
+	for _, tx := range c.Message.Transactions {
+		txHash := sha256.Sum256(*tx)
+		hasher.Write(txHash[:])
+	}
+
+	return hasher.Sum(nil)
+}
+
+func (c *SignedConstraints) VerifySignature(pubkey phase0.BLSPubKey) (bool, error) {
+	signingData := phase0.SigningData{ObjectRoot: phase0.Root(c.Digest()), Domain: phase0.Domain([]byte{109, 109, 111, 67})}
```
merklefruit: can we add a comment explaining the COMMIT_BOOST_DOMAIN?
maybe saving it into a variable as well would be neat

builder/common/types.go
```diff
@@ -597,40 +616,88 @@ func (v *VersionedSubmitBlockRequestWithProofs) String() string {
...
+	hasher.Write(d.Message.DelegateePubkey[:])
+	return hasher.Sum(nil)
 }
 
-func (c *ConstraintSubscriptionAuth) String() string {
-	buf, err := json.Marshal(c)
+// VerifySignature verifies the signature of a signed delegation. IMPORTANT: it uses the COMMIT_BOOST_DOMAIN
+// as the signing domain: [109, 109, 111, 67]
+func (d *SignedDelegation) VerifySignature(pubkey phase0.BLSPubKey) (bool, error) {
+	signingData := phase0.SigningData{ObjectRoot: phase0.Root(d.Digest()), Domain: phase0.Domain([]byte{109, 109, 111, 67})}
```
merklefruit: same as above

builder/miner/worker.go
```diff
@@ -1106,68 +1115,77 @@ func (w *worker) commitTransactions(env *environment, plainTxs, blobTxs *transac
...
+			common.Filter(&constraintsBySender, func(txRecovered *types.TransactionEcRecovered) bool {
+				return txRecovered.Sender == from
+			})
+
+			lowestNonceConstraintBySender := common.Last(constraintsBySender)
+			if lowestNonceConstraintBySender.Transaction.Nonce() < tx.Nonce() {
+				// This means that the constraint with the lowest nonce from this sender
+				// has lower nonce than the pooled tx, so we cannot execute the pooled tx yet.
+				// We need to execute the constraint first.
+				candidate = candidateTx{tx: lowestNonceConstraintBySender.Transaction, isConstraint: true}
```
merklefruit, line 1172: This is a great change to the worker algo, I am trying to wrap my head around the diffs here but I like it.
@thedevbirb can you confirm this is basically handling the edge case where there are 2 transactions one in the mempool and one in constraints and they must be ordered according to their nonce?

merklefruit (changes requested):

builder/builder/builder.go
mempirate: You're right, we should not only move on to the next pubkey but if there is no valid signature, the constraint should be discarded. Addressed in latest commit

builder/builder/builder.go
mempirate: Don't see a clear reason to either remove the proposer key or keep it in the list, but since it's a list I suppose we can keep it. To me it seemed more logical that if a proposer decides to delegate to another key, its own key shouldn't be used anymore for signing constraints. Wdyt?

merklefruit: I agree, it's fine as it is then. I still suggest documenting this behaviour in the API specs / docs

builder/common/types.go
mempirate: Added a bit more explanation + fixes in latest commits. Will document also in the API docs!

builder/builder/builder.go
```diff
@@ -395,7 +411,20 @@ func (b *Builder) Stop() error {
 	return nil
 }
 
-// BOLT: modify to calculate merkle inclusion proofs for preconfirmed transactions
+// GetConstraintsDomain returns the constraints domain used to sign constraints-API related messages.
+//
+// The builder signing domain is built as follows:
+// - We build a ForkData ssz container with the fork version and the genesis validators root. In the builder domain, this is an empty root.
```
thedevbirb, line 417: What do you mean with "In the builder domain, this is an empty root"?

builder/builder/relay_aggregator.go
```diff
@@ -81,6 +81,56 @@ func (r *RemoteRelayAggregator) SubmitBlockWithProofs(msg *common.VersionedSubmi
...
+
+			delegationsCh <- &delegations
+		}(relay, i)
+	}
+
+	err := errors.New("could not get delegations from any relay")
+
+	aggregated := types.SignedDelegations{}
+	for i := 0; i < len(r.relays); i++ {
+		d := <-delegationsCh
```
thedevbirb, line 104: Is there a timeout for the `GetDelegationsForSlot`? Otherwise this might hang waiting for the some relays to respond.

builder/miner/worker.go
thedevbirb: @merklefruit the builder algorithm simply handles constraints by putting them at the bottom of the block, however sometimes this may violate protocol rules if there is a transaction in the mempool with the same sender and an higher nonce. That happens if I send a transaction in the mempool after I've sent a constraint.

The pooled transaction would error with "Nonce too high" and would be discarded. This check takes care of it, by looking at constraints made from that sender and placing them before the pooled transaction if necessary.

Hope it is clear! If you want me to re-word the comments lmk

thedevbirb (commented):
Great work @mempirate, the changes you made are clear to me. Just two minor questions

builder/builder/builder.go
mempirate: Across all implementations, the builder signing domain is built with an empty genesis validators root. I don't know why. In contrast, the beacon signing domain always contains a valid gvr.

builder/miner/worker.go
merklefruit: Got it, awesome thanks for explaining. All clear

builder/builder/builder.go
merklefruit: If it can be helpful, here is more context on the decision:

https://github.com/ethereum/builder-specs/issues/14#issuecomment-1127055996

> With that said, it makes the most sense to me to compute the domain for an application using `compute_domain(APP_DOMAIN_TYPE, fork_version=None, genesis_validators_root=None)`, because I don't think application should need any external data to compute / verify a signature.

mempirate: The builder domain does use the `fork_version` field though. I think for us that also makes sense to ensure signatures are only valid for the fork they're running on. Wdyt?

merklefruit (approved):
LGTM

PR #251, opened 2024-10-01
feat(registry): added max committed gas limit per validator

### LINKED ISSUE #250: feat(contracts): add per-validator gas limit in registry

merklefruit:
* added missing docs
* removed `Validator.authorizedCollateralProvider` because it wasn't used
* added `Validator.maxCommittedGasLimit`

Note: this PR adds this value to the registry + middleware contracts, but it doesn't update the Challenger or any of the off-chain components reading the registry such as the RPC. That will be done in separate PRs.

- closes #250


mempirate (approved):
LGTM!

PR #256, opened 2024-10-02
fix: update types & drop bundle support

### LINKED ISSUE #217: Drop support for bundle preconfs
## Context
Bundle preconfs have various negative externalities. Until we figure those out, remove support.

EDIT: we could also loosen bundle specifications by saying they must be executed with an ascending index, but not continuous.

mempirate:
Closes #217

### DIFF
--- a/bolt-sidecar/src/driver.rs
+++ b/bolt-sidecar/src/driver.rs
@@ -220,21 +220,23 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS, 
 
         // parse the request into constraints and sign them
         let slot = inclusion_request.slot;
-        let message = ConstraintsMessage::build(validator_pubkey, inclusion_request);
-        let signed_constraints = match self.constraint_signer.sign(&message.digest()).await {
-            Ok(signature) => SignedConstraints { message, signature },
-            Err(err) => {
-                error!(?err, "Failed to sign constraints");
-                let _ = response.send(Err(CommitmentError::Internal));
-                return;
-            }
-        };
 
-        // Track the number of transactions preconfirmed considering their type
-        signed_constraints.message.transactions.iter().for_each(|full_tx| {
-            ApiMetrics::increment_transactions_preconfirmed(full_tx.tx_type());
-        });
-        self.execution.add_constraint(slot, signed_constraints);
+        for tx in inclusion_request.txs {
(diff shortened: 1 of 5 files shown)


bolt-sidecar/src/driver.rs
```diff
@@ -220,21 +220,26 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
...
+                Ok(signature) => SignedConstraints { message, signature },
+                Err(err) => {
+                    error!(?err, "Failed to sign constraints");
+                    let _ = response.send(Err(CommitmentError::Internal));
+                    return;
+                }
+            };
+
+            ApiMetrics::increment_transactions_preconfirmed(tx_type);
+            self.execution.add_constraint(slot, signed_constraints);
```
thedevbirb, line 241:
```suggestion
            let signed_constraint = match self.constraint_signer.sign(&message.digest()).await {
                Ok(signature) => SignedConstraint { message, signature },
                Err(err) => {
                    error!(?err, "Failed to sign constraint");
                    let _ = response.send(Err(CommitmentError::Internal));
                    return;
                }
            };

            ApiMetrics::increment_transactions_preconfirmed(tx_type);
            self.execution.add_constraint(slot, signed_constraint);
```
For clarity that is `signed_constraintS -> `signed_constraint`, also for the type. Can we drop also the `Vec` from `SignedConstraints`? Or it is used elsewhere.

thedevbirb, line 241: Does this work if a transaction appears twice, first time in a commitment request with just one tx and the second time in a commitment with a batch of two transactions?
Because there is a risk otherwise we increment the nonce of balance of the user two times

thedevbirb (commented):
Two little q's!

bolt-sidecar/src/driver.rs
mempirate: If we change it from a `Vec` that would change the constraints-API. We don't use constraint bundles as of now but we might still use it in the future. We just dropped support for bundles from the commitments-API

merklefruit (approved):

bolt-sidecar/src/driver.rs
mempirate: I created a quick test with a duplicate tx in a batch, which should and does fail with nonce too low. It's not exactly what you described but should cover it.

thedevbirb (approved):

PR #265, opened 2024-10-04
fix: remove metrics clash with cb client

merklefruit:
bolt-boost in kurtosis had some panics due to already-registered metrics. disambiguated by adding `_bolt`.


thedevbirb (approved):
gr8

PR #266, opened 2024-10-04
fix: insert into constraints cache

merklefruit:
the `ConstraintsCache` was getting stuck in a branch waiting for a second write lock.
Fixed and added a test including a check for conflicting transactions


mempirate (approved):
Good catch

PR #267, opened 2024-10-04
chore(ci): add bolt_boost ci + run only if there are diffs on its path

merklefruit:
Now the actions will run only if there are diffs on the specified paths.


thedevbirb (approved):
Nice, thank you!

PR #275, opened 2024-10-07
fix(sidecar): commit boost flag rename

thedevbirb:
Imo the existing `commit_boost_address` flag doesn't convey properly that an URL should be passed, so the flag has been renamed.

### DIFF
--- a/bolt-sidecar/src/config/mod.rs
+++ b/bolt-sidecar/src/config/mod.rs
@@ -111,7 +111,7 @@ pub struct Config {
     /// The engine API url
     pub engine_api_url: Url,
     /// URL for the commit-boost sidecar
-    pub commit_boost_address: Option<String>,
+    pub commit_boost_url: Option<String>,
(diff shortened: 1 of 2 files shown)


bolt-sidecar/src/client/commit_boost.rs
```diff
@@ -39,8 +39,8 @@ pub enum CommitBoostError {
 #[allow(unused)]
 impl CommitBoostSigner {
     /// Create a new [CommitBoostSigner] instance
-    pub async fn new(signer_server_address: String, jwt: &str) -> Result<Self, CommitBoostError> {
-        let signer_client = SignerClient::new(signer_server_address, jwt)?;
+    pub async fn new(signer_server_address: &str, jwt: &str) -> Result<Self, CommitBoostError> {
```
mempirate: Nit but usually the function interface should show if it requires an owned or borrowed variable, i.e. it shouldn't do a hidden clone

mempirate (approved):
lgtm

PR #278, opened 2024-10-07
feat(contracts): registry updates

### LINKED ISSUE #233: feat(contracts): add `ChainSpec` to `BeaconChainUtils`
The BeaconChainUtils helper contract clearly relies on some hard-coded values for Ethereum Mainnet.
Before deploying to any testnet, we should parametrize this contract to work on more chains such as Holesky / Helder.

TODOs:

- [ ] make BeaconChainUtils parametric over chain id
- [ ] implement the chain spec for a few networks we plan to work with

Another option is to have different copies of the contract such as `MainnetBeaconChainUtils` but I think it's less clean.

### LINKED ISSUE #252: feat(contracts): add RPC in operator registration
## Context
Operators should register their RPC along with their signing key so consumers of the registry know where to send requests.

### LINKED ISSUE #273: Holesky deployment plan
Image

mempirate:
- Closes #252 
- Closes #273


.gitmodules
```diff
@@ -13,3 +13,12 @@
 [submodule "bolt-contracts/lib/eigenlayer-contracts"]
 	path = bolt-contracts/lib/eigenlayer-contracts
 	url = https://github.com/layr-labs/eigenlayer-contracts
+[submodule "lib/openzeppelin-contracts-upgradeable"]
+	path = lib/openzeppelin-contracts-upgradeable
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
+[submodule "bolt-contracts/lib/openzeppelin-contracts-upgradeable"]
+	path = bolt-contracts/lib/openzeppelin-contracts-upgradeable
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
```
merklefruit: this seems like a duplicate module, can you check which one we're actually using?

bolt-contracts/src/contracts/BoltEigenLayerMiddleware.sol
```diff
@@ -165,43 +152,39 @@ contract BoltEigenLayerMiddleware is IBoltMiddleware, Ownable {
...
+        // Register the operator in the manager
+        boltManager.registerOperator(operator, rpc);
     }
 
-    /// @notice Allow an operator to signal indefinite opt-out from Bolt Protocol.
-    /// @dev Pausing activity does not prevent the operator from being slashable for
-    /// the current network epoch until the end of the slashing window.
-    function pauseOperator() public {
-        if (!operators.contains(msg.sender)) {
+    /// @notice Deregister an EigenLayer layer operator from working in Bolt Protocol.
```
merklefruit:
```suggestion
    /// @notice Deregister an EigenLayer operator from working in Bolt Protocol.
```

bolt-contracts/src/contracts/BoltManager.sol
merklefruit, line 1: symbiotic contract imports at lines 10-17 can be removed (they are leftovers from previous versions)

bolt-contracts/src/contracts/BoltManager.sol
```diff
@@ -33,20 +34,59 @@ contract BoltManager is IBoltManager, Ownable {
...
-    // ========= CONSTRUCTOR =========
+    /// @notice Start timestamp of the first epoch.
+    uint48 public START_TIMESTAMP;
+
+    // ============= CONSTANTS ================= //
+    /// @notice Duration of an epoch in seconds.
+    uint48 public constant EPOCH_DURATION = 1 days;
+
+    modifier onlyMiddleware() {
+        require(restakingProtocols.contains(msg.sender), "BoltManager: caller is not a middleware");
```
merklefruit: nit: so far we've been sticking to named error messages specified in interfaces. Can we use one here too instead of a `require` statement? for consistency

bolt-contracts/src/contracts/BoltManager.sol
```diff
@@ -66,21 +106,134 @@ contract BoltManager is IBoltManager, Ownable {
...
+
+        uint48 epochStartTs = getEpochStartTs(getEpochAtTs(Time.timestamp()));
+        IBoltValidators.Validator memory validator = validators.getValidatorByPubkeyHash(pubkeyHash);
+
+        Operator memory operator = operators.get(validator.authorizedOperator);
+
+        status.pubkeyHash = pubkeyHash;
+        status.active = validator.exists;
+        status.operator = validator.authorizedOperator;
+        status.operatorRPC = operator.rpc;
```
merklefruit: I like this a lot!

merklefruit (changes requested):
small nits, overall I love the design and how much it simplifies our current middlewares. great!

bolt-contracts/src/contracts/BoltParameters.sol
thedevbirb, line 1: Just one performance consideration: we're already concerned about the gas usage of the challenger and while some of these variables should be upgradable, some of them instead are pretty much known and immutable once the contract has been deployed.
In particular those are the ones that don't require an admin method:
- `EPOCH_DURATION`
- `SLASHING_WINDOW`
- `ETH2_GENESIS_TIMESTAMP`
- `SLOT_TIME`.
These are used in the Challenger contract via a static call that consumes more gas than it should. I'd personally prefer rely on those via a library whose constant values are set manually before deploying. Ugly yes, but we can get around it with a script reminding us to do that before deploying.

bolt-contracts/src/contracts/BoltManager.sol
```diff
@@ -66,21 +119,154 @@ contract BoltManager is IBoltManager, Ownable {
...
+
+    /// @notice Allow an operator to signal indefinite opt-out from Bolt Protocol.
+    /// @dev Pausing activity does not prevent the operator from being slashable for
+    /// the current network epoch until the end of the slashing window.
+    function pauseOperator(
+        address operator
+    ) external onlyMiddleware {
+        if (!operators.contains(operator)) {
+            revert OperatorNotRegistered();
+        }
```
thedevbirb: I don't think this is a necessary step since `disable` already reverts on a missing key. See https://github.com/chainbound/bolt/blob/jonas%2Ffeat%2Fregistry-updates/bolt-contracts/src/lib/EnumerableMap.sol#L49-L55. 
I get it is more explicit, it is just a gas cost consideration. You can ignore this if you think it's not relevant.

bolt-contracts/src/contracts/BoltManager.sol
```diff
@@ -66,21 +119,154 @@ contract BoltManager is IBoltManager, Ownable {
...
+        operators.disable(operator);
+    }
+
+    /// @notice Allow a disabled operator to signal opt-in to Bolt Protocol.
+    function unpauseOperator(
+        address operator
+    ) external onlyMiddleware {
+        if (!operators.contains(operator)) {
+            revert OperatorNotRegistered();
+        }
```
thedevbirb: Same here and for other occurrences of `operators.contains`.

bolt-contracts/test/BoltChallenger.t.sol
```diff
@@ -51,7 +56,32 @@ contract BoltChallengerTest is Test {
...
 
+        uint48 epochDuration = 1 days;
+        uint48 slashingWindow = 7 days;
+        uint48 maxChallengeDuration = 7 days;
+        bool allowUnsafeRegistration = true;
+        uint256 challengeBond = 1 ether;
+        uint256 blockhashEvmLookback = 256;
+        uint256 justificationDelay = 32;
+        uint256 eth2GenesisTimestamp = 1_606_824_023;
+        uint256 slotTime = 12;
```
thedevbirb: These are copy-pasted many times, any chance we can source it from a JSON file? I think Nico has done it in the past

bolt-contracts/test/BoltManager.EigenLayer.t.sol
```diff
@@ -160,35 +200,35 @@ contract BoltManagerEigenLayerTest is Test {
...
-        uint256 totalStake = middleware.getTotalStake(2, address(eigenLayerDeployer.weth()));
-        assertEq(amount, 1 ether);
-        assertEq(totalStake, 1 ether);
-    }
+    //     uint256 amount = middleware.getOperatorStake(operator, address(eigenLayerDeployer.weth()));
+    //     // TODO:
+    //     uint256 totalStake = middleware.getTotalStake(2, address(eigenLayerDeployer.weth()));
+    //     assertEq(amount, 1 ether);
+    //     assertEq(totalStake, 1 ether);
+    // }
```
thedevbirb: Is this intentional? Or can you write why you cannot do it now?

bolt-contracts/test/BoltManager.Symbiotic.t.sol
```diff
@@ -282,9 +315,10 @@ contract BoltManagerTest is Test {
         assertEq(vault.currentEpoch(), 2);
 
         // it takes 2 epochs to activate the stake
-        assertEq(middleware.getTotalStake(0, address(collateral)), 0);
-        assertEq(middleware.getTotalStake(1, address(collateral)), 0);
-        assertEq(middleware.getTotalStake(2, address(collateral)), 1 ether);
+        // TODO:
+        // assertEq(middleware.getTotalStake(0, address(collateral)), 0);
+        // assertEq(middleware.getTotalStake(1, address(collateral)), 0);
+        // assertEq(middleware.getTotalStake(2, address(collateral)), 1 ether);
```
thedevbirb: Is this intentional? Or can you write why you cannot uncomment them now?

testnets/holesky/.cb.env
thedevbirb, line 1: Can these be safely exposed?

testnets/holesky/cb.docker-compose.yml
thedevbirb, line 1: Why is this file in this PR?

thedevbirb (commented):
Amazing work! Just some notes

bolt-contracts/src/contracts/BoltParameters.sol
merklefruit: One note is that gas usage in the challenger is way less of a concern than the registry.

I think this is fine as is for this use case.

fwiw, there is much lower hanging fruit for optimizing the opening / resolution process than this, so I suggest leaving it here as it makes the whole deployment simpler!

bolt-contracts/src/contracts/BoltChallenger.sol
```diff
@@ -22,26 +30,41 @@ contract BoltChallenger is IBoltChallenger {
 
     // ========= STORAGE =========
 
+    /// @notice Bolt Parameters contract.
+    IBoltParameters public parameters;
+
     /// @notice The set of existing unique challenge IDs.
     EnumerableSet.Bytes32Set internal challengeIDs;
```
merklefruit, line 39: note: when counting storage layout, each variable should be counted for the amount of storage it uses.
In this case, `EnumerableSet.Bytes32Set` is defined as `Set` under the hood:

```solidity
    struct Set {
        bytes32[] _values;
        mapping(bytes32 value => uint256) _positions;
    }
```

this occupies 2 slots: 1 for the array (pointer to its mem address) and 1 for the mapping.
reference: https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays

so TL;DR: storage layout marker would be 4 instead of 3 slots here. I didn't check other contracts but this is a good check to make everywhere

bolt-contracts/src/contracts/BoltManager.sol
```diff
@@ -1,52 +1,105 @@
...
+
     /// @notice Set of restaking protocols supported. Each address corresponds to the
     /// associated Bolt Middleware contract.
     EnumerableSet.AddressSet private restakingProtocols;
 
-    // ========= CONSTRUCTOR =========
+    // --> Storage layout marker: 4 slots
+
+    /// @notice Start timestamp of the first epoch.
+    uint48 public START_TIMESTAMP;
```
merklefruit: same comment as above. plus, why isn't START_TIMESTAMP counted in the storage layout here?

bolt-contracts/src/contracts/BoltParameters.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    /// @notice The number of slots to wait before considering a block justified by LMD-GHOST.
+    uint256 public JUSTIFICATION_DELAY;
+
+    /// @notice The timestamp of the eth2 genesis block.
+    uint256 public ETH2_GENESIS_TIMESTAMP;
+
+    /// @notice The duration of a slot in seconds.
+    uint256 public SLOT_TIME;
+    // --> Storage layout marker: 6 words
```
merklefruit: This is good, but to make sure it is correct I suggest using solc json storage layout outputs.
ref: https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-input-and-output-json-description
and https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#json-output

merklefruit (changes requested):
Looking very good!

bolt-contracts/src/contracts/BoltParameters.sol
mempirate: Good point. My thinking was that other contracts might also need these values, but we can have a `BoltConstants` library per network (mainnet, holesky, helder, ...) that we can just use in all those contracts. Will update!

bolt-contracts/src/contracts/BoltManager.sol
mempirate: Because it doesn't take up a full storage slot yet. In case other variables smaller than 256 - 48 need to be added they can be added until the slot is full, after which we need to decrement the gap.

bolt-contracts/src/contracts/BoltChallenger.sol
mempirate: Good catch

bolt-contracts/src/contracts/BoltManager.sol
merklefruit: It is still occupying the 5th slot though; If other variables smaller than 256 - 48 get added, they will share the 5th slot with it. That's what I understood

bolt-contracts/test/BoltChallenger.t.sol
mempirate: Done in latest commit! Required a bit of scaffolding but lmk what you think

bolt-contracts/src/contracts/BoltManager.sol
mempirate: Yeah you're right, fixed

bolt-contracts/src/contracts/BoltParameters.sol
mempirate: This is sick: `forge inspect BoltParameters storage-layout --pretty`

bolt-contracts/src/contracts/BoltChallenger.sol
mempirate: Validated all storage layouts with `forge inspect BoltParameters storage-layout --pretty`

Added instructions on how to do it in the contract natspec documentation. The OpenZeppelin foundry upgrades toolkit can also validate it every time we do a new deployment.

bolt-contracts/src/contracts/BoltParameters.sol
merklefruit: TIL!

merklefruit (approved):
LGTM

PR #281, opened 2024-10-07
feat(sidecar): sign messages using the correct signing domain

merklefruit:
Signing and verifying messages with BLS keys now complies to the consensus-layer signing domains spec.
In particular, two signing domains are provided: `APPLICATION_BUILDER` and `COMMIT_BOOST`.

Signers are able to sign() or verify() messages in both domains explicitly to avoid incurring in inconsistencies.

Note: the beacon-chain specs also define signing over SSZ containers via hash_tree_root. We currently don't do that, and have decided to use our own signing digests that are simpler and don't require deriving SSZ. Since this is an out-of-protocol API we are not required to comply to this restriction.

### DIFF
--- a/bolt-sidecar/src/client/commit_boost.rs
+++ b/bolt-sidecar/src/client/commit_boost.rs
@@ -109,11 +109,11 @@ impl CommitBoostSigner {
 
 #[async_trait::async_trait]
 impl SignerBLS for CommitBoostSigner {
-    async fn sign(&self, data: &[u8; 32]) -> eyre::Result<BlsSignature> {
-        let request = SignConsensusRequest::builder(
-            *self.pubkeys.read().first().expect("consensus pubkey loaded"),
-        )
-        .with_msg(data);
+    async fn sign_commit_boost_root(&self, data: &[u8; 32]) -> eyre::Result<BlsSignature> {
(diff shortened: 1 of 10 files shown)


bolt-sidecar/src/client/commit_boost.rs
```diff
@@ -109,11 +109,11 @@ impl CommitBoostSigner {
...
-    async fn sign(&self, data: &[u8; 32]) -> eyre::Result<BlsSignature> {
-        let request = SignConsensusRequest::builder(
-            *self.pubkeys.read().first().expect("consensus pubkey loaded"),
-        )
-        .with_msg(data);
+    async fn sign_commit_boost_root(&self, data: &[u8; 32]) -> eyre::Result<BlsSignature> {
+        let request = SignConsensusRequest {
+            pubkey: *self.pubkeys.read().first().expect("consensus pubkey loaded"),
+            object_root: *data,
+        };
```
merklefruit, line 116: `with_msg()` was calling `tree_hash_root()` internally. Since we already have a 32 bytes message, we don't need to do this. If a tree hash root is required, it must be hashed by the caller.

mempirate (approved):
lgtm!

PR #276, opened 2024-10-07
feat(mev-boost): update constraints api

namn-grg:
Closes #272

PR #285, opened 2024-10-08
fix: updated devnet

### LINKED ISSUE #284: Revert disabling authorized pubkeys once testing is done on devnet
Revert https://github.com/chainbound/bolt/pull/282. Requires having clear how to handle delegations in the devnet, and possibly #283.

merklefruit:
Assorted fixes for the devnet before Holesky.

### Constraints API

* added `action` flag to the Delegation and Revocation messages, to disambiguate between the two
* associated docs PR: https://github.com/chainbound/bolt-docs/pull/22

### Sidecar

* streamlined `BlsPublicKey` types to use less versions of the type (we have 4 different crates exporting it)
* sign constraints inserting the signer pubkey instead of the target slot's validator pubkey, to unblock testing. This will require more attention when we want to make sure to sign with a key that is either authorized via delegation or is the validator private key itself
* added tests w.r.t. constraints encoding, decoding & sig verify
* applied changes to the constraints api for delegation and revocation messages

### Builder

* fix: decode constraints as single message instead of lists when unmarshaling them from SSE
* fix: runtime error in `commitTransaction` implementation
* fix: missing `BlobsBundle` in deneb submit block request MarshalJSON implementation
* added tests w.r.t. constraints unmarshaling
* re-added checks to verify that constraints are submitted by authorized keys

### Bolt Boost

* fix: calculate proof data correctly for type 3 blob txs
* applied changes to the constraints api for delegation and revocation messages

### Mev Boost

* applied changes to the constraints api for delegation and revocation messages

### Meta

* closes #284
* reverted #282
* added #287


bolt-boost/src/types.rs
```diff
@@ -92,14 +101,45 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
             .transactions
             .iter()
             .map(|tx| {
-                let tx_hash = *TxEnvelope::decode_2718(&mut tx.as_ref())?.tx_hash();
+                let envelope = TxEnvelope::decode_2718(&mut tx.as_ref())?;
+                let tx_hash = *envelope.tx_hash();
```
thedevbirb: Nit but can we move this into a separate utility function? One that takes raw bytes and returns a `(TxHash, TxEnvelope)` tuple. It'd be used also above for digest calculation.

bolt-boost/src/types.rs
```diff
@@ -92,14 +101,45 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
...
+                        }
+                    },
+                    // For other transaction types, take the hash tree root of the whole tx
+                    _ => tree_hash::TreeHash::tree_hash_root(&Transaction::<
                         <DenebSpec as EthSpec>::MaxBytesPerTransaction,
-                    >::from(tx.to_vec()));
+                    >::from(
+                        tx.to_vec()
+                    )),
+                };
```
thedevbirb: Same here, this can be a separate utility function that calculates the hash tree root for a `TxEnvelope`.

thedevbirb (commented):
~I tried to rebase this PR but didn't work, I still see commits from #276. Can you try to update it?~ 

Nvm, I thought #276 was merged into `unstable`.

thedevbirb (approved):
daje

PR #287, opened 2024-10-10
feat(sidecar): ERC-2335 Keystore signer

### LINKED ISSUE #283: Support per-validator signer
## Context
Right now the sidecar accepts one constraint signer as a CLI param for all of its validators. This won't work because the relay now considers the only valid constraint submission key to be the validator pubkey by default. The sidecar should read the keys from an [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335) compliant keystore (I think all VCs support this). This will provide a mapping between validator pubkeys and private keys.

This targets validators that are not using an external signer (Web3Signer, DIRK), and also don't care about delegating to a different key.

thedevbirb:
Closes #283. This PR introduces support for running the sidecar by reading the private keys from ERC-2335 keystore files.

Such keystore are expected to be placed inside the `keys` folder of this project which reserved for this purpose only. While we could allow fetching them from a specified path, I like the idea of the sidecar not reading arbitrary folders of the users by mistake.

Some refactoring changes have been done as well:
- All different signers have been moved to a `signer` folder
- `SignerBLS` trait has been transformed into an enum because there's not a common trait for all type of signers
- `Config` struct has been absorbed into `Opts` by squeezing `Clap` functionalities

Relies on https://github.com/chainbound/ethereum-package/pull/17.


bolt-sidecar/keys/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!.gitignore
+!README.md
```
merklefruit, line 3: pro level stuff right here :)

bolt-sidecar/src/config/mod.rs
```diff
@@ -22,255 +17,78 @@ pub use signing::SigningOpts;
...
-    #[clap(long, env = "BOLT_SIDECAR_PORT")]
-    pub(super) port: Option<u16>,
+    #[clap(long, env = "BOLT_SIDECAR_PORT",
+        default_value_t = DEFAULT_RPC_PORT)]
+    pub port: u16,
     /// URL for the beacon client
-    #[clap(long, env = "BOLT_SIDECAR_BEACON_API_URL")]
-    pub(super) beacon_api_url: String,
+    #[clap(long, env = "BOLT_SIDECAR_BEACON_API_URL",
+        default_value_t = Url::parse("http://localhost:5052").expect("Valid Beacon API URL"))]
```
merklefruit: I think it's cleaner to do `default_value = "http://localhost:5052"` if it compiles

bolt-sidecar/src/config/mod.rs
```diff
@@ -22,255 +17,78 @@ pub use signing::SigningOpts;
...
+    pub constraints_proxy_port: u16,
     /// Validator indexes of connected validators that the sidecar
     /// should accept commitments on behalf of. Accepted values:
     /// - a comma-separated list of indexes (e.g. "1,2,3,4")
     /// - a contiguous range of indexes (e.g. "1..4")
     /// - a mix of the above (e.g. "1,2..4,6..8")
-    #[clap(long, value_parser = ValidatorIndexes::from_str, env = "BOLT_SIDECAR_VALIDATOR_INDEXES")]
-    pub(super) validator_indexes: ValidatorIndexes,
+    #[clap(long, value_parser = ValidatorIndexes::from_str, env = "BOLT_SIDECAR_VALIDATOR_INDEXES",
+        default_value_t = ValidatorIndexes::default())]
```
merklefruit: clap(default) should do the trick if I'm not mistaken (also everywhere the default() method is used)

bolt-sidecar/src/config/signing.rs
```diff
@@ -1,29 +1,133 @@
...
+}
+
+// Implement Debug manually to hide the keystore_password field
+impl fmt::Debug for SigningOpts {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SigningOpts")
+            .field("private_key", &self.private_key)
+            .field("commit_boost_url", &self.commit_boost_address)
+            .field("commit_boost_jwt_hex", &self.commit_boost_jwt_hex)
+            .field("keystore_password", &"********") // Hides the actual password
```
merklefruit, line 42: nice

bolt-sidecar/src/config/signing.rs
```diff
@@ -1,29 +1,133 @@
...
+            keystore_password: None,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BlsSecretKey(pub SecretKey);
+
+impl BlsSecretKey {
+    pub fn random_bls_secret() -> Self {
```
merklefruit: can you rename this to just `random`?
Also this type could probably be moved to common.rs

bolt-sidecar/src/config/signing.rs
```diff
@@ -1,29 +1,133 @@
...
+}
+
+impl fmt::Display for BlsSecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "0x{}", hex::encode(self.0.to_bytes()))
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct JwtSecretConfig(pub String);
```
merklefruit: I'm pretty sure reth has a similar type already, maybe worth checking if it's suitable to our case

bolt-sidecar/src/driver.rs
```diff
@@ -58,76 +60,99 @@ impl<B: SignerBLS> fmt::Debug for SidecarDriver<StateClient, B, PrivateKeySigner
...
+        constraint_signer: SignerBLS,
         commitment_signer: ECDSA,
         fetcher: C,
     ) -> eyre::Result<Self> {
-        let constraints_client = ConstraintsClient::new(cfg.constraints_url.clone());
-        let beacon_client = BeaconClient::new(cfg.beacon_api_url.clone());
-        let execution = ExecutionState::new(fetcher, cfg.limits).await?;
+        let constraints_client = ConstraintsClient::new(opts.constraints_url.clone());
+        let beacon_client = BeaconClient::new(opts.beacon_api_url.clone());
+        let execution = ExecutionState::new(fetcher, opts.limits.clone()).await?;
```
merklefruit: I see that you removed the derive(Copy) from Limits, is there a reason or can you add it back?

bolt-sidecar/src/signer/keystore.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub const KEYSTORES_DEFAULT_PATH: &str = "keys";
+
+#[derive(Clone)]
+pub struct KeystoreSigner {
+    keypairs: Vec<Keypair>,
+}
+
+impl KeystoreSigner {
+    pub fn new(keys_path: Option<&str>, password: &[u8]) -> eyre::Result<Self> {
```
merklefruit: Shouldn't the password be optional? I'm not familiar with the EIP in particular.

bolt-sidecar/src/signer/keystore.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        Ok(Self { keypairs })
+    }
+
+    pub fn sign_commit_boost_root(
+        &self,
+        root: [u8; 32],
+        public_key: [u8; BLS_PUBLIC_KEY_BYTES_LEN],
+    ) -> eyre::Result<BLSSig> {
+        // NOTE: is this a good comparison?
```
merklefruit: hmm, one thing that comes to mind could be casing, e.g. 0xaBc vs 0xABc. 
if you can do the check on bytes that would be perfect, otherwise maybe simply setting to lowercase could work

bolt-sidecar/src/signer/local.rs
```diff
@@ -0,0 +1,132 @@
...
+            Ok(())
+        } else {
+            eyre::bail!(format!("Invalid signature: {:?}", res))
+        }
+    }
+}
+
+/// Generate a random BLS secret key.
+#[cfg(test)]
+pub fn random_bls_secret() -> BlsSecretKey {
```
merklefruit: I believe you now have this defined in the signing config, can you use that?

bolt-sidecar/src/signer/mod.rs
```diff
@@ -0,0 +1,19 @@
+use keystore::KeystoreSigner;
+use local::LocalSigner;
+
+use crate::CommitBoostSigner;
+
+pub mod commit_boost;
+pub mod keystore;
+pub mod local;
```
merklefruit:
```suggestion
use crate::CommitBoostSigner;

pub mod keystore;
use keystore::KeystoreSigner;

pub mod local;
use local::LocalSigner;

pub mod commit_boost;
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -13,7 +13,7 @@ use tracing::{debug, trace};
 use crate::{
     builder::BlockTemplate,
     common::{calculate_max_basefee, max_transaction_cost, validate_transaction},
-    config::Limits,
+    config::limits::LimitsOpts as Limits,
```
merklefruit: why do we need this alias?

bolt-sidecar/src/test_util.rs
```diff
@@ -75,25 +78,32 @@ pub(crate) async fn try_get_beacon_api_url() -> Option<&'static str> {
 /// - The default values for the remaining configuration fields.
 ///
 /// If any of the above values can't be found, the function will return `None`.
-pub(crate) async fn get_test_config() -> Option<Config> {
+pub(crate) async fn get_test_config() -> Option<Opts> {
+    let sk = BlsSecretKey::random_bls_secret().to_string();
+    println!("sk: {}", sk);
```
merklefruit:
```suggestion
```

merklefruit (changes requested):
The config refactor is very nice! Left some comments but overall great.

Wrt the signer, this is what I understand from the current flow with the Keystore:
* sidecar is started with keystore CLI flags + `validator-indexes` flag
* at each slot, we check if the proposer duties contain one of our validator indexes
* if so, we try to find the pubkey from the keystore and sign the constraints with it

It would be cleaner to use the keystore as source of truth for which pubkeys we can use to sign messages in the first place, instead of going through `validator_indexes`.

In particular here's how I imagined it:
* sidecar is started with keystore CLI flags
* all pubkeys are loaded in memory when creating the keystore
* at each slot, we check if the proposer duties contain on of our pubkeys
* if so, simply sign it with that key

let me know your thoughts

bolt-sidecar/src/config/mod.rs
thedevbirb: Good call yes, it compiles and it is less verbose.

bolt-sidecar/src/config/mod.rs
thedevbirb: Actually I discovered that using `default_value_t` without specifying such value does the trick. Neat!

bolt-sidecar/src/config/signing.rs
thedevbirb: Yes I've checked it before implementing it and it is a wrapper over `[u8; 32]` but the underlying array is not accessible, and it doesn't implement `ToString` or `Default`. As such I could not transform it in a hex string as needed by Commit-Boost.

merklefruit: in alloy it has `as_bytes` but still not sure if worth to use it. maybe take a look https://github.com/alloy-rs/alloy/blob/ef9b58b617e33eba326a219a4dfb705b72db0dde/crates/rpc-types-engine/src/jwt.rs#L266

bolt-sidecar/src/test_util.rs
thedevbirb: Yeah sorry about that, still fixing some tests. Going to address it later

bolt-sidecar/src/signer/keystore.rs
thedevbirb: Unfortunately there isn't really a `to_bytes()` method, but I've found a `as_ssz_bytes()` method that maybe can work. I'll use that and see it that works.

bolt-sidecar/src/signer/keystore.rs
thedevbirb: According to the [ERC](https://eips.ethereum.org/EIPS/eip-2335#password-verification) and the [lighthouse implementation](https://github.com/sigp/lighthouse/blob/stable/crypto/eth2_keystore/src/keystore.rs#L250) it is required. It is just a better practice imo, if you want to avoid a password you can always use `password` :)

thedevbirb:
> The config refactor is very nice! Left some comments but overall great.
> 
> Wrt the signer, this is what I understand from the current flow with the Keystore:
> 
> * sidecar is started with keystore CLI flags + `validator-indexes` flag
> * at each slot, we check if the proposer duties contain one of our validator indexes
> * if so, we try to find the pubkey from the keystore and sign the constraints with it
> 
> It would be cleaner to use the keystore as source of truth for which pubkeys we can use to sign messages in the first place, instead of going through `validator_indexes`.
> 
> In particular here's how I imagined it:
> 
> * sidecar is started with keystore CLI flags
> * all pubkeys are loaded in memory when creating the keystore
> * at each slot, we check if the proposer duties contain on of our pubkeys
> * if so, simply sign it with that key
> 
> let me know your thoughts

Thanks for the great review! Yes the flow you described is indeed correct. You're right that in the context of the keystore signer the process of checking validator indexes is redundant. I have to think whether it's still needed for other type of signers though

bolt-sidecar/src/signer/keystore.rs
merklefruit: Got it thanks

bolt-sidecar/src/signer/keystore.rs
merklefruit, line 1: How annoying would it be to add an `KeystoreError` enum using thiserror?

bolt-sidecar/src/test_util.rs
```diff
@@ -1,3 +1,4 @@
+use crate::ChainConfig;
```
merklefruit: can you bring this import down closer to his frens after `use crate::`? it's sad and alone here

bolt-sidecar/src/config/signing.rs
thedevbirb: Oh I see, haven't found it because it was implemented recently (2024-09-25) with this commit: https://github.com/alloy-rs/alloy/commit/6ff6a9d91a7004f529eb1f2dc6735e7d3420ef69

bolt-sidecar/src/driver.rs
```diff
@@ -224,24 +248,41 @@ impl<C: StateFetcher, BLS: SignerBLS, ECDSA: SignerECDSA> SidecarDriver<C, BLS,
         // parse the request into constraints and sign them
         let slot = inclusion_request.slot;
 
+        let pubkey = match self.constraint_signer {
```
merklefruit, line 251: reminder: we're missing a check here. this pubkey is used to sign constraints but we don't know if it is authorized to do so. this can be added after we load delegation info in #290

bolt-sidecar/src/signer/commit_boost.rs
```diff
@@ -34,14 +30,19 @@ pub enum CommitBoostError {
     #[error("failed to sign constraint: {0}")]
     NoSignature(String),
     #[error("failed to create signer client: {0}")]
-    SignerClientError(#[from] ErrReport),
+    SignerClientError(#[from] SignerClientError),
+    #[error("error in commit boost signer: {0}")]
+    Other(#[from] eyre::Report),
 }
 
+type Result<T> = std::result::Result<T, CommitBoostError>;
```
merklefruit: in general I'm not a big fan of renaming std library objects, it's confusing. I understand you did this to avoid replacing all instances of the word, but I'm sure this is just 1 vim macro away :)

I would simply rename this to `CBResult` or `CommitBoostResult`. then you also don't need to qualify std::result::Result anymore.

same for other signer results

merklefruit (approved):
Great job!

bolt-sidecar/src/signer/commit_boost.rs
thedevbirb: I have seen this comment only now, I wasn't 100% happy with the current errors either and I've made a commit that should resolve your comment as well: [`ddaedb5` (#287)](https://github.com/chainbound/bolt/pull/287/commits/ddaedb52c89dce8604a0d36570dc9ac21d85c9ea). wdyt?

merklefruit: yep awesome, ty

PR #292, opened 2024-10-14
feat(sidecar): load delegations on startup and send them upon registration

### LINKED ISSUE #290: feat(sidecar): Create a delegations file format and load it on startup
We may need to load existing delegations on startup, and there currently isn't a format to do so.
Proposed solution: 

A JSON file with the following format: `Vec<SignedDelegation>` to use our existing types.

```json
[
  {
    "message": {
      "action": 0,
      "validator_pubkey": "0x...",
      "delegatee_pubkey": "0x..."
    },
    "signature": "0x..."
  },
  ...
]
```

At startup, the sidecar will attempt to load these delegations, and then it will cross-check the delegated pubkeys with the ones available in the provided signer (local, remote or via keystore files).

then the sidecar has all the info necessary to sign constraint messages when needed.

merklefruit:
## Overview

* Added the ability to parse `Opts` from a TOML file, loaded optionally if the `BOLT_SIDECAR_CONFIG_PATH` is defined
* Added an example toml config validated through a unit test (so it will be easy to keep it updated over time)
* Added the `signing.delegations` file path option that will parse SignedDelegations from JSON if present
* Added logic for checking if a delegated public key is available before signing constraint messages
* Added call to `delegate()` during `register_validators()` so they are done together to signal delegations regularly. We should maybe think of a better strategy as this could be inefficient if done with many delegations (e.g. 10k+ validators).
* Refactored some primitives (e.g. extracted `delegation` and `transaction` types to their own file).

Note: this PR changes the constraints API body type for `delegate()` and `revoke()` to take Vec<T> instead of T.
Following PRs in bolt-boost and mev-boost should reflect this change.

## Meta

* closes #290


bolt-sidecar/Config.toml
thedevbirb, line 1: What is the usage of this? Or is it an example configuration file?

bolt-sidecar/src/driver.rs
```diff
@@ -245,13 +273,18 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
...
+        // Pick a pubkey to sign constraints with.
+        //
+        // Rationale:
+        // - If there are no delegatee keys, try to use the validator key directly if available.
+        // - If there are delegatee keys, try to use the first one that is available in the list.
+        let Some(pubkey) = pick_public_key(validator_pubkey, available_pubkeys, delegatees) else {
+            error!(%target_slot, "No available public key to sign constraints with");
+            let _ = response.send(Err(CommitmentError::Internal));
+            return;
         };
```
merklefruit: this is quite inefficient but the logic should check out. Let me know your thoughts!

bolt-sidecar/src/config/mod.rs
thedevbirb, line 1: Just to be sure can you run `just fmt bolt-sidecar`? I see some inconsistencies here.

bolt-sidecar/src/config/mod.rs
```diff
@@ -90,16 +89,43 @@ pub struct Opts {
...
         let parsed = url.parse::<Url>().unwrap();
         let socket_addr = parsed.socket_addrs(|| None).unwrap()[0];
         let localhost_socket = "0.0.0.0:3030".parse().unwrap();
         assert_eq!(socket_addr, localhost_socket);
     }
+
+    #[test]
+    fn test_parse_config_from_toml() {
+        let path = env!("CARGO_MANIFEST_DIR").to_string() + "Config.toml";
+        dbg!(&path);
```
thedevbirb:
```suggestion
```

bolt-sidecar/src/config/signing.rs
```diff
@@ -30,6 +29,9 @@ pub struct SigningOpts {
     /// Path to the keystores folder. If not provided, the default path is used.
     #[clap(long, env = "BOLT_SIDECAR_KEYSTORE_PATH", requires("keystore_password"))]
     pub keystore_path: Option<String>,
+    /// Path to the delegations file. If not provided, the default path is used.
+    #[clap(long, env = "BOLT_SIDECAR_DELEGATIONS_PATH")]
+    pub delegations_path: Option<String>,
```
thedevbirb, line 34: Can you add it to the `Debug` implementation?

bolt-sidecar/src/driver.rs
```diff
@@ -245,13 +273,18 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
...
-            SignerBLS::CommitBoost(ref signer) => signer.pubkey(),
-            SignerBLS::Keystore(_) => validator_pubkey.clone(),
+        let delegatees = self.constraints_client.find_delegatees(&validator_pubkey);
+        let available_pubkeys = self.constraint_signer.available_pubkeys();
+
+        // Pick a pubkey to sign constraints with.
+        //
+        // Rationale:
+        // - If there are no delegatee keys, try to use the validator key directly if available.
+        // - If there are delegatee keys, try to use the first one that is available in the list.
```
thedevbirb: The `pick_public_key` function is already documented, if you want you can drop this!

bolt-sidecar/src/driver.rs
```diff
@@ -354,3 +390,29 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
...
+
+/// Pick a pubkey to sign constraints with.
+///
+/// Rationale:
+/// - If there are no delegatee keys, try to use the validator key directly if available.
+/// - If there are delegatee keys, try to use the first one that is available in the list.
+fn pick_public_key(
+    validator: BlsPublicKey,
+    available: Vec<BlsPublicKey>,
+    delegatees: Vec<BlsPublicKey>,
```
thedevbirb: Any chance we can make these `HashSet<BlsPublicKey>` compared to `Vec`? So we don't need to loop over every time, and this should also address your concerns about the logic being a bit inefficient.

bolt-sidecar/src/signer/keystore.rs
```diff
@@ -57,6 +59,17 @@ impl KeystoreSigner {
...
 
+    /// Returns the public keys of the keypairs in the keystore.
+    pub fn pubkeys(&self) -> Vec<BlsPublicKey> {
+        self.keypairs
+            .iter()
+            .map(|kp| {
+                BlsPublicKey::try_from(kp.pk.serialize().to_vec().as_ref()).expect("valid pubkey")
+            })
+            .collect::<Vec<_>>()
+    }
```
thedevbirb: Related to the comment above, this should be an easy change
```suggestion
    /// Returns the public keys of the keypairs in the keystore.
    pub fn pubkeys(&self) -> HashSet<BlsPublicKey> {
        self.keypairs
            .iter()
            .map(|kp| {
                BlsPublicKey::try_from(kp.pk.serialize().to_vec().as_ref()).expect("valid pubkey")
            })
            .collect::<HashSet<_>>()
    }
```

thedevbirb (commented):
Overall it's great! Some q's

bolt-sidecar/Config.toml
merklefruit: it's an example, also used by a unit test to check integrity.
Maybe I should also include the options that have a default value for completeness?

thedevbirb: As you wish. Can you just rename it to `Config.example.toml`? Just to be sure this is not used as a default somewhere but mainly as a reference example or for test

bolt-sidecar/src/driver.rs
merklefruit: good point

bolt-sidecar/src/signer/keystore.rs
merklefruit: yep, done as suggested above

bolt-sidecar/delegations.json
mempirate, line 1: Should we move this to testdata?

mempirate (approved):
Great work, lgtm

bolt-sidecar/src/signer/keystore.rs
mempirate: 1 quick point, for small collections a Vec is generally faster than a HashSet for iterating and finding apparently: https://gist.github.com/daboross/976978d8200caf86e02acb6805961195?permalink_comment_id=4400446#gistcomment-4400446

thedevbirb (approved):

bolt-sidecar/delegations.json
merklefruit: done

PR #293, opened 2024-10-15
feat: add delegations cli

### LINKED ISSUE #286: Tool for signing a delegation message
## Context

In case validators want to sign constraints with another key, they need to sign a delegation message that can be used by the sidecar to signal a delegation downstream. The sidecar can then use the delegated key to sign any further constraints, omitting the need for the validator key to be used more than once.

## Implementation

We need to create a CLI tool (let's call it `bolt-delegations-cli` for now, we can always change it). This is a Rust CLI tool that runs with the following command:

```
bolt-delegations-cli generate --source keystore ./path/to/keystore --delegated-key 0x1234 --out delegations.json
```

* We can start with a single `generate` subcommand that takes the following options:
  * --source: can be either `local` or `keystore` (required)
    * if `local`, read a BLS private key as the next argument
    * if `keystore`, read a file path as the next argument
  * --delegated-key: a BLS pubkey to which the delegation messages should be signed to (required)
  * --out: the location where to output the delegations file (can default to delegations.json)

The goal of this is to be a fully offline tool that node operators will be able to use to generate delegations safely and airgap their bolt sidecar from validator client or sensitive keystore information.

### Details

* we should re-use part of the keystore implementation made by @thedevbirb in #287 
* we can also take some inspiration from the [`go-eth2-wallet`](https://github.com/wealdtech/go-eth2-wallet?tab=readme-ov-file#signing-data-and-verifying-signatures) tool

namn-grg:
# Overview
- Added a simple cli tool to generate signed delegation message.
- The private key for signing the message can be read from :
  - A local private key saved in a file for example `private_key.txt` 
  - A keystore (taken help from #287 )

# Meta
closes #286 

related PR #292


bolt-delegations-cli/.gitignore
thedevbirb, line 1: Can you also add `target/`, `.env.*`, `!.env.example`?. Those values should be in the root of the Bolt project  as well but just to be sure

bolt-delegations-cli/README.md
```diff
@@ -0,0 +1,65 @@
+# Bolt Delegations CLI
+
+`bolt-delegations-cli` is a command-line tool for generating delegation messages signed with a BLS (Boneh–Lynn–Shacham) key. This tool allows node operators to safely generate delegations offline, enabling an air-gapped workflow to secure sensitive information like validator keys.
```
thedevbirb:
```suggestion
`bolt-delegations-cli` is an offline command-line tool for safely generating delegation messages signed with a BLS12-381 key for the [Constraints API](https://docs.boltprotocol.xyz/api/builder) in [Bolt](https://docs.boltprotocol.xyz/).
```
We're using BLS12-381 which is actually made by Barreto, Lynn and Scott.

bolt-delegations-cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+The tool supports two key sources:
+
+-   Local: A BLS private key provided directly from a file.
+-   Keystore: A keystore file that contains an encrypted BLS private key, with the set [default password](https://github.com/chainbound/bolt/blob/a935fb36d75c997a4edb834f27a56bc62eb3570c/bolt-delegations-cli/src/utils.rs#L11).
```
thedevbirb: The password should not be set by default, that is a canonical test case for the ERC-2335. Can you reflect this also in the rest of the README?

bolt-delegations-cli/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+/// - Decrypt the keypair using the default password (TODO: make this configurable)
+/// - Create a delegation message
+/// - Compute the signing root and sign the message
+/// - Return the signed delegation
+fn generate_from_keystore(
+    key_path: &str,
+    delegatee_pubkey: BlsPublicKey,
+    chain: &Chain,
+) -> Result<SignedDelegation> {
+    let keypair = Keystore::from_json_file(key_path)
```
merklefruit: note: the key_path should be the `directory` holding the keypairs, not just a single file. Like it's done here: https://github.com/chainbound/bolt/blob/unstable/bolt-sidecar/src/signer/keystore.rs#L45

this means that the `delegations.json` output file must be a serialized `Vec<SignedDelegation>`, not just a single one. We want to create a delegation from all available validator keys to a single delegatee key.

bolt-delegations-cli/src/lib.rs
merklefruit, line 1: since this is a binary, we don't need to expose a lib for now.
can you delete this and move these lines to main.rs instead?

bolt-delegations-cli/src/config.rs
```diff
@@ -0,0 +1,62 @@
...
+pub enum Commands {
+    /// Generate delegation messages.
+    Generate {
+        /// The source of the validator key (local or keystore).
+        #[clap(long, env = "SOURCE")]
+        source: SourceType,
+
+        /// Path to the keystore file or private key, depending on the source.
+        #[clap(long, env = "KEY_PATH")]
+        key_path: String,
```
thedevbirb: As suggested by @mempirate also, usually private keys are not stored in plain text. You can take inspiration from the `SigningOpts` in the sidecar and make the user choose between a path to the keystore, along with its password, or a private key passed directly.

bolt-delegations-cli/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    let signing_root = compute_signing_root_for_delegation(&delegation, chain)?;
+    let sig = sk.sign(signing_root.0.as_ref());
+
+    Ok(SignedDelegation { message: delegation, signature: sig })
+}
+
+/// Generate a signed delegation using a keystore file
+///
+/// - Read the keystore file
+/// - Decrypt the keypair using the default password (TODO: make this configurable)
```
thedevbirb: I see there is a TODO for it, do you want to address it in a separate PR for a specific reason?

bolt-delegations-cli/src/types.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/// Event types that can be emitted by the validator pubkey to
+/// signal some action on the Bolt protocol.
+#[derive(Debug, Clone, Copy)]
+#[repr(u8)]
+enum SignedMessageAction {
+    /// Signal delegation of a validator pubkey to a delegatee pubkey.
+    Delegation,
+}
```
thedevbirb: Can you add `Revocation` as well? Ideally this CLI should be used for those as well.

thedevbirb (changes requested):
Thanks for the PR! Can you also add some tests? You can take a look at `tempfile` crate if you need to create temporary files and drop them when the test end.

bolt-delegations-cli/README.md
merklefruit: BLS wars

bolt-delegations-cli/src/main.rs
namn-grg: we discussed in standup, I'll do in this pr only

bolt-delegations-cli/src/lib.rs
namn-grg: yes!

bolt-delegations-cli/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+2. Using an Ethereum keystore file:
+
+    ```bash
+    bolt-delegations-cli generate-keystore \
+        --keystore-path /keys \
+        --keystore-password myS3cr3tP@ssw0rd \
+        --delegatee-pubkey 0x7890ab... \
+        --out my_delegations.json \
+        --chain kurtosis
```
thedevbirb: I like the subcommands as done now, can you update this readme as well?

thedevbirb (approved):

bolt-delegations-cli/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+
+                write_to_file(&out, &messages)?;
+                println!("Signed delegation messages generated and saved to {}", out);
+            }
+        },
+    }
+    Ok(())
+}
+
+/// Generate signed delegations/devocations using local BLS private keys
```
thedevbirb:
```suggestion
/// Generate signed delegations/revocations using local BLS private keys
```

bolt-delegations-cli/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+                signed_messages.push(SignedMessage::Delegation(signed))
+            }
+            Action::Revoke => {
+                let message = RevocationMessage::new(sk.public_key(), delegatee_pubkey.clone());
+                let signing_root = compute_commit_boost_signing_root(message.digest(), chain)?;
+                let signature = sk.sign(signing_root.0.as_ref());
+                let signed = SignedRevocation { message, signature };
+                signed_messages.push(SignedMessage::Revocation(signed));
+            }
+        }
```
thedevbirb, line 89:
```suggestion
        let message = DelegationMessage::new(sk.public_key(), delegatee_pubkey.clone());
        let signing_root = compute_commit_boost_signing_root(message.digest(), chain)?;
        let signature = sk.sign(signing_root.0.as_ref());
        let signed = SignedDelegation { message, signature };
        
        match action {
            Action::Delegate => {
                signed_messages.push(SignedMessage::Delegation(signed))
            }
            Action::Revoke => {
                signed_messages.push(SignedMessage::Revocation(signed));
            }
        }
```

bolt-delegations-cli/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            Action::Revoke => {
+                let message = RevocationMessage::new(validator_pubkey, delegatee_pubkey.clone());
+                let signing_root = compute_commit_boost_signing_root(message.digest(), chain)?;
+                let signature = validator_private_key.sign(signing_root.0.into());
+                let signature = BlsSignature::try_from(signature.serialize().as_ref())?;
+                let signed = SignedRevocation { message, signature };
+                signed_messages.push(SignedMessage::Revocation(signed));
+            }
+        }
```
thedevbirb: Can be simplified in a similar fashion as above

bolt-delegations-cli/src/types.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub message: RevocationMessage,
+    pub signature: BlsSignature,
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
+pub struct RevocationMessage {
+    action: u8,
+    pub validator_pubkey: BlsPublicKey,
+    pub delegatee_pubkey: BlsPublicKey,
+}
```
thedevbirb, line 91: Do we need this distinction also in the codebase?

thedevbirb (changes requested):
Minor things!

bolt-delegations-cli/src/main.rs
merklefruit: this won't work because `DelegationMessage` and `RevocationMessage` are different types

bolt-delegations-cli/src/main.rs
merklefruit: same as above, it can't - the best we could do is add a `sign_message()` function that takes in a digest, to decrease the duplication going on, however it's also a bit annoying because the `GenericSecretKey<...>` type from lighthouse is not very portable (looks like we would need to import other lighthouse libraries to expose it directly... uff)

bolt-delegations-cli/src/types.rs
merklefruit: it's really only useful when creating the messages, and we have it there.
for transporting (e.g. bolt-boost, mev-boost) it won't matter

bolt-delegations-cli/src/main.rs
merklefruit: oopsie

thedevbirb (approved):

Issue #295, opened 2024-10-16
chore(sidecar): review default options for CLI flags


thedevbirb:
There are many configuration options in the sidecar that have a default (for testing reasons) but they shouldn't have one, and we should make sure the users manually sets them. Some examples include: 
- fee recipient
- builder private key
- jwt hex
- etc..

This requires also modifying our fork of the Ethereum package in order to provide the correct values and run the devnet successfully.

A PR should be built on top of #296 which already touches some of the CLI flag descriptions etc.

PR #300, opened 2024-10-18
chore(sidecar): default options for cli flag

namn-grg:
Removes default options for the following CLI flag -
- fee recipient
- builder private key
- jwt hex

solves #295


merklefruit:
incorporated in #301

PR #301, opened 2024-10-18
feat(sidecar): add commitment signing key + holesky deployment scripts

merklefruit:
* supersedes #300 
* adds a commitment signing key mandatory flag to the sidecar


bolt-sidecar/src/config/constraint_signing.rs
```diff
@@ -6,21 +6,21 @@ use serde::Deserialize;
...
-    #[clap(long, env = "BOLT_SIDECAR_PRIVATE_KEY")]
-    pub private_key: Option<BlsSecretKeyWrapper>,
+pub struct ConstraintSigningOpts {
+    /// Private key to use for signing constraint messages
+    #[clap(long, env = "BOLT_SIDECAR_CONSTRAINT_PRIVATE_KEY")]
+    pub constraint_private_key: Option<BlsSecretKeyWrapper>,
     /// Socket address for the commit-boost sidecar
     #[clap(long, env = "BOLT_SIDECAR_CB_SIGNER_URL", requires("commit_boost_jwt_hex"))]
-    pub commit_boost_address: Option<SocketAddr>,
+    pub commit_boost_signer_url: Option<SocketAddr>,
```
thedevbirb: This name can be slightly misleading since we require a socket address and not a full URL. There are two approaches:
1. rename this to `commit_boost_signer_socket_address` -- long but does the job
2. expect a full URL and then parse it to extract just the socket address

bolt-sidecar/src/config/mod.rs
```diff
@@ -46,10 +46,18 @@ pub struct Opts {
     #[clap(long, env = "BOLT_SIDECAR_ENGINE_API_URL", default_value = "http://localhost:8551")]
     pub engine_api_url: Url,
     /// URL for the Constraint sidecar client to use
```
thedevbirb: Since we're touching the flags, what do you think of this description?

```suggestion
    /// URL to forward the constraints produced by the Bolt sidecar to a server supporting the Constraints API, such as an MEV-Boost fork.
```

bolt-sidecar/src/config/mod.rs
```diff
@@ -46,10 +46,18 @@ pub struct Opts {
...
-    #[clap(long, env = "BOLT_SIDECAR_CONSTRAINTS_URL", default_value = "http://localhost:3030")]
-    pub constraints_url: Url,
+    #[clap(
+        long,
+        env = "BOLT_SIDECAR_CONSTRAINTS_API_URL",
+        default_value = "http://localhost:3030"
+    )]
+    pub constraints_api_url: Url,
     /// Constraint proxy server port to use
```
thedevbirb: Since we're touching the flags, what do you think of this description?

```suggestion
    /// The port from which the Bolt sidecar will receive Builder-API requests from the Beacon client.
```

bolt-sidecar/src/driver.rs
```diff
@@ -90,13 +90,17 @@ impl SidecarDriver<StateClient, PrivateKeySigner> {
...
+                .clone()
+                .expect("local constraint signing key")
+                .0,
             opts.chain,
         ));
 
         // Commitment responses are signed with a regular Ethereum wallet private key.
         // This is now generated randomly because slashing is not yet implemented.
-        let commitment_signer = PrivateKeySigner::random();
+        let commitment_signer = PrivateKeySigner::from_slice("".as_bytes().to_vec().as_slice())?;
```
thedevbirb: Why are we initialising the `PrivateKeySigner` from an empty string?

justfile
```diff
@@ -172,20 +172,20 @@ _build-bolt-boost:
...
-deploy-sidecar-dev:
-    chmod +x ./scripts/deploy_bolt_sidecar.sh && ./scripts/deploy_bolt_sidecar.sh
+deploy-sidecar-dev chain:
+    chmod +x ./scripts/deploy_bolt_sidecar.sh && ./scripts/deploy_bolt_sidecar.sh {{chain}}
 
 # Check the status of the sidecar service on the dev server
-status-sidecar-dev:
-    ssh shared@remotebeast "sudo systemctl status bolt_sidecar" | less
+status-sidecar-dev chain:
+    ssh shared@remotebeast "sudo systemctl status bolt_sidecar{{chain}}" | less
```
thedevbirb: Probably a typo here

```suggestion
    ssh shared@remotebeast "sudo systemctl status bolt_sidecar_{{chain}}" | less
```

thedevbirb (changes requested):
Great, some minor things

bolt-sidecar/src/config/constraint_signing.rs
merklefruit: done with (2)

bolt-sidecar/src/driver.rs
merklefruit: sorry was a placeholder

justfile
merklefruit: ty!

thedevbirb (approved):

PR #302, opened 2024-10-18
feat(contracts): script to register validators

merklefruit:


merklefruit:
closed as part of #202

PR #304, opened 2024-10-21
fix: symbiotic network opt-in

merklefruit:
Added a check to make the operator opt in to the network before registering in the middleware.
This will be skipped if it's already done by the operator beforehand.

### DIFF
--- a/bolt-contracts/config/holesky/deployments.json
+++ b/bolt-contracts/config/holesky/deployments.json
@@ -3,13 +3,13 @@
         "validators": "0x05275a4799cd1B07D81319390fC62Bc7BDbDf269"
     },
     "symbiotic": {
-        "network": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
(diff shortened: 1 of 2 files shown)


bolt-contracts/config/holesky/deployments.json
```diff
@@ -19,7 +19,7 @@
         "avsDirectory": "0x055733000064333CaDDbC92763c58BF0192fFeBf",
         "delegationManager": "0xA44151489861Fe9e3055d95adC98FbD462B948e7",
         "strategyManager": "0xdfB5f6CE42aAA7830E94ECFCcAd411beF4d4D5b6",
-        "middleware": "",
+        "middleware": "0x279F5be63FD1D0A32d77258028933a7e3E75fF08",
```
mempirate: Let's keep this out of the config so we don't accidentally set incorrect addresses

mempirate (approved):
I like it - not going to do the same for EigenLayer since it requires more information like operator metadata etc.

PR #306, opened 2024-10-21
Document on-chain registration

mempirate:
Provides the steps required for the on-chain registration of Validators and Operators in the Holesky guide.


thedevbirb (approved):
Thanks for the help!

Issue #307, opened 2024-10-22
`bolt-cli` revamp


mempirate:
## Context
Because we'll add a lot more functionality to the CLI, we should rename the `bolt-delegations-cli` to just `bolt-cli` (with binary command just being `bolt`). We'll also want to leverage it for the following:

```[tasklist]
# Commands
- [x] Integrate with DIRK
- [x] Querying Bolt-related information on-chain
- [x] send transactions (replace `bolt-client` and `bolt-kurtosis-client`)
- [ ] https://github.com/chainbound/bolt/issues/365
- [ ] https://github.com/chainbound/bolt/issues/392
- [x] Easily register operators (replace foundry scripts + better configuration)
```

PR #309, opened 2024-10-22
feat: `bolt-cli` revamp: add Dirk support

merklefruit:
starts #307 

## Overview

This PR adds support for Dirk as remote signer for generating BLS key delegations.
We can now list, unlock remote accounts and sign arbitrary data through the eth2-signer-api.

This PR comes with an e2e test using Dirk as a background process with pre-fabricated certificates and wallets.
To run the test, use the following command:

```shell
 cargo test --package bolt-cli --bin bolt-cli -- utils::dirk::tests::test_dirk_connection_e2e --exact --show-output --ignored
```

The guide can be found in the [README here](https://github.com/chainbound/bolt/tree/nico/feat/bolt-cli/bolt-cli).

tasks:

- [x] update docs
- [x] more tests
- [x] auto re-lock accounts after signing


bolt-cli/Cargo.toml
```diff
@@ -1,30 +1,42 @@
 [package]
-name = "bolt-delegations-cli"
+name = "bolt-cli"
```
mempirate: Can we rename this to just `bolt`? This will be the canonical CLI package, users would just have to invoke `bolt operator register ...` etc

thedevbirb: Agreed! Please remember to change the invocations of `bolt-cli` to `bolt` in the README then

thedevbirb (approved):
Apart from the renaming the code looks great! It's awesome there is also a test in the CI, great work.

bolt-cli/README.md
```diff
@@ -0,0 +1,213 @@
...
+2. Generating a delegation using an ERC-2335 keystore directory
+
+```text
+bolt-cli delegate \
+ --delegatee-pubkey 0x8d0edf4fe9c80cd640220ca7a68a48efcbc56a13536d6b274bf3719befaffa13688ebee9f37414b3dddc8c7e77233ce8 \
+ --chain holesky \
+ local-keystore --path test_data/lighthouse/validators --password-path test_data/lighthouse/secrets
+```
+
+3. Generating a revocation using a remote DIRK keystore
```
mempirate: Shouldn't this be a delegation?

bolt-cli/Cargo.toml
merklefruit: done!

bolt-cli/README.md
merklefruit: fixed

PR #310, opened 2024-10-23
feat(holesky): updated EL scripts and README

thedevbirb:
Updates the EigenLayer scripts and README to opt in as an operator.


testnets/holesky/README.md
```diff
@@ -559,13 +594,14 @@ sidecar. You can see them in your terminal by running the Bolt sidecar binary
...
+
 Command-line options for the Bolt sidecar
 
 Usage: bolt-sidecar [OPTIONS] --validator-indexes <VALIDATOR_INDEXES> --engine-jwt-hex <ENGINE_JWT_HEX> --fee-recipient <FEE_RECIPIENT> --builder-private-key <BUILDER_PRIVATE_KEY> --commitment-private-key <COMMITMENT_PRIVATE_KEY> <--constraint-private-key <CONSTRAINT_PRIVATE_KEY>|--commit-boost-signer-url <COMMIT_BOOST_SIGNER_URL>|--keystore-password <KEYSTORE_PASSWORD>|--keystore-secrets-path <KEYSTORE_SECRETS_PATH>>
 
 Options:
-      --port <PORT>
-          Port to listen on for incoming JSON-RPC requests of the Commitments API. This port should be open on your firewall in order to receive external requests!
+--port <PORT>
+Port to listen on for incoming JSON-RPC requests of the Commitments API. This port should be open on your firewall in order to receive external requests!
```
merklefruit: I think this formatting is wrong, can you re-add this spacing? also below

merklefruit (approved):

bolt-contracts/config/holesky/operator.json
```diff
@@ -1,5 +1,6 @@
 {
-    "rpc": "localhost:50051",
-    "salt": "0x000000000000000abc0000000000000000000000000000000000000000000000",
-    "expiry": null
-}
\ No newline at end of file
+  "rpc": "<hostname>:<port>",
+  "salt": "0x0000000000000000000_salt_value_0000000000000000000000000000000000",
```
mempirate: This can be anything that is a bytes32 (so a hash or something), maybe we should add a section on what the salt does and how one can generate it using bash commands, options:
- `head -c 32 /dev/urandom | xxd -p -c 64`
- `openssl rand -hex 32`

(output will have to be 0x prefixed)

bolt-contracts/config/holesky/operator.json
```diff
@@ -1,5 +1,6 @@
...
+  "expiry": "0x00000000000000000_expiry_value_0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
mempirate: Same here. Expiry is the hex-encoding of a UNIX timestamp in seconds. Can convert it to hex with `printf "%x\n" $TIMESTAMP_DECIMALS`

bolt-contracts/script/holesky/operators/RegisterEigenLayerOperator.s.sol
```diff
@@ -19,9 +20,27 @@ contract RegisterEigenLayerOperator is Script {
...
+
+        vm.startBroadcast(operatorSk);
+        // Allowance must be set before depositing
+        token.approve(address(strategyManager), amount);
+        strategyManager.depositIntoStrategy(strategy, token, amount);
+        console.log("Successfully run StrategyManager.depositIntoStrategy");
+        vm.stopBroadcast();
+    }
+
+    function S02_registerIntoBoltAVS() public {
```
mempirate, line 42: Can we remove the `S0_1 prefixes on the functions, we don't have it anywhere else. Unless you have a good reason

mempirate (changes requested):
Nice job, requested some smol changes

bolt-contracts/script/holesky/operators/RegisterEigenLayerOperator.s.sol
thedevbirb: The reasons are the following:
- having a single file for each script would resulting in copying over the same utilities functions over and over again, so I put everything into one file, `RegisterEigenLayerOperator.s.sol`
- given this scripts, it makes to ask whether there is an order with which you should run them. This allows you to self-document the script file and I took inspiration from it from Forge's scripts best practices https://book.getfoundry.sh/tutorials/best-practices#scripts

testnets/holesky/README.md
thedevbirb: Thanks for noticing!

bolt-contracts/script/holesky/operators/RegisterEigenLayerOperator.s.sol
mempirate: Got it, then let's leave it!

PR #313, opened 2024-10-23
feat(holesky): update README and scripts for Symbiotic integration

thedevbirb:
To be merged after #310


merklefruit (approved):

bolt-contracts/script/holesky/operators/RegisterSymbioticOperator.s.sol
```diff
@@ -14,33 +17,41 @@ contract RegisterSymbioticOperator is Script {
         address symbioticNetwork;
     }
 
-    function run() public {
+    function S01_registerIntoBolt() public {
         uint256 operatorSk = vm.envUint("OPERATOR_SK");
+        string memory rpc = vm.envString("OPERATOR_RPC");
```
mempirate, line 21: Why don't we read this from the config file? Or did we get rid of it (`operator.json`)

mempirate: We did

PR #316, opened 2024-10-24
chore!(sidecar,holesky): drop support for TOML, use .env everywhere

### LINKED ISSUE #308: Drop `.env` file support and keep only TOML configuration file
Currently we support both that this results in extra work and an higher chance to have stale `.example` files with can result in a worse UX. Also TOML is more flexible compared to an `.env` file.

thedevbirb:
Closes #308  by doing the opposite: use `.env` files everywhere because of consistency with other services + better support via Clap.


testnets/holesky/README.md
```diff
@@ -875,10 +884,10 @@ bolt --version
 The binary can be used with the following command:
 
 ```shell
-bolt delegate --delegate-pubkey <DELEGATEE_PUBKEY> 
-              --out <OUTPUT_FILE> 
-              --chain <CHAIN> 
-              <KEY_SOURCE> 
+bolt delegate --delegate-pubkey <DELEGATEE_PUBKEY>
```
merklefruit: typo I made earlier, my bad
```suggestion
bolt delegate --delegatee-pubkey <DELEGATEE_PUBKEY>
```

merklefruit (approved):
I agree this feels a lot cleaner, nice job

PR #321, opened 2024-10-25
feat(cli): `bolt send` command

merklefruit:
Refactor our CLI packages in order to use `bolt send` to send a preconfirmation request.

part of #307


bolt-cli/src/commands/send.rs
```diff
@@ -0,0 +1,159 @@
...
+        };
+
+        let lookahead_res = reqwest::get(lookahead_url).await?.json::<Vec<LookaheadSlot>>().await?;
+        if lookahead_res.is_empty() {
+            println!("no bolt proposer found in the lookahead, try again later 🥲");
+            return Ok(());
+        }
+
+        // Extract the next preconfirmer slot from the lookahead info
+        let target_slot = lookahead_res[0].slot;
```
thedevbirb, line 68: Can we do `.get(0).expect("...")` instead?

thedevbirb (approved):

bolt-cli/src/commands/send.rs
merklefruit: we check if it's empty in the line above, so that would be unreachable

mempirate (approved):
LGTM!

PR #329, opened 2024-10-28
feat(ci): add linkspector

mempirate:


.linkspector.yml
merklefruit, line 1: is it possible to place this inside `.github/` such as to not clutter the top-level?
Not a big problem of course but we're already keeping all the CI-related stuff here

merklefruit (approved):
LGTM

thedevbirb:
This is cool, can we add it also for the `testnets` or `testnets/holesky` folder?

PR #330, opened 2024-10-28
`enable-unsafe-lookahead` flag

thedevbirb:
Introduces a new `enable-unsafe-lookahead` flag in configuration that allows to validate commitment requests along two consecutive epochs instead of just the current one.

### DIFF
--- a/bolt-sidecar/src/state/consensus.rs
+++ b/bolt-sidecar/src/state/consensus.rs
@@ -5,6 +5,7 @@ use std::{
 
 use beacon_api_client::{mainnet::Client, ProposerDuty};
 use ethereum_consensus::{crypto::PublicKey as BlsPublicKey, phase0::mainnet::SLOTS_PER_EPOCH};
+use tokio::join;
 use tracing::debug;
 
 use super::CommitmentDeadline;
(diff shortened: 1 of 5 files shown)


bolt-sidecar/src/state/consensus.rs
```diff
@@ -58,6 +64,8 @@ pub struct ConsensusState {
     pub commitment_deadline: CommitmentDeadline,
     /// The duration of the commitment deadline.
     commitment_deadline_duration: Duration,
+    /// If commitment requests should be validated against also against the unsafe lookahead
```
mempirate: Lil typo

bolt-sidecar/src/state/consensus.rs
```diff
@@ -101,14 +111,18 @@ impl ConsensusState {
     ) -> Result<BlsPublicKey, ConsensusError> {
         let CommitmentRequest::Inclusion(req) = request;
 
-        // Check if the slot is in the current epoch
-        if req.slot < self.epoch.start_slot || req.slot >= self.epoch.start_slot + SLOTS_PER_EPOCH {
+        let furthest_slot = self.epoch.start_slot
```
mempirate: Can we extract this into a helper like `get_furthest_slot`

```rs
fn get_furthest_slot() -> u64 {
  if unsafe_lookahead_enabled {
    self.epoch.start_slot + SLOTS_PER_EPOCH * 2
  } else {
    self.epoch.start_slot + SLOTS_PER_EPOCH
  }
} 
```

mempirate (changes requested):
Great job! Just a couple of suggestions

mempirate (approved):
🚀

PR #335, opened 2024-10-28
docs(holesky): clarify order of operations

mempirate:


testnets/holesky/README.md
```diff
@@ -88,596 +90,601 @@ or **authorized delegates** acting on their behalf, to issue and sign preconfirm
...
+# On-Chain Registration
 
-- Docker mode (recommended)
-- [Commit-Boost](https://commit-boost.github.io/commit-boost-client) mode
-  (requires Docker)
-- Native mode (advanced, requires building everything from source)
+Once you are successfully running the Bolt sidecar you need to register on-chain
+on the Bolt Registry to successfully receive preconfirmation requests from users
+and RPCs. This step is needed to provide economic security to your
+commitments.
```
thedevbirb: I think this paragraph isn't consistent with the previous callout: you're previously asking to run the on-chain steps first, but then you say _"Once you are successfully running the Bolt sidecar..."_ which suggests you should run the sidecar first.

Wdyt?

mempirate: True, will take another look at that

thedevbirb (approved):

PR #336, opened 2024-10-28
feat(registry): cheaper on-chain batch opt-in (48k gas)

### LINKED ISSUE #338: Add necessary parameters to errors in smart contracts
Currently some errors in `BoltValidatorsV1` like `ValidatorDoesNotExist` or `ValidatorAlreadyExists` do not take any argument which makes hard to debug which validator public key is actually not correct, especially in batch calls like `getProposerStatuses`

Ideally all similar errors should be parametric.

merklefruit:
## Overview

This PR introduces a slightly more optimized version of `BoltValidators` that uses an internal library `ValidatorsLib`.

The main gain in efficiency is due to the new `_Validator` struct that tightly packs necessary info in a single slot, containing pointers to other mappings, offloading the majority of storage costs when batch registering.

The optimizations performed are geared toward storage costs rather than runtime costs.
In particular the loops, variables, event emission, and other runtime checks haven't been optimized/removed.

## Gas usage

We benchmarked `BoltValidatorsV2` (new implementation) vs `BoltValidatorsV1` (old implementation).

* previous version: ~120k gas x validator (when batching). Max 250 per block gas limit.
* new version: ~48k gas x validator (when batching). Max 620 per block gas limit

The image below shows the efficiency curve of batch registration costs:

Gas Cost vs Amount of Validators (Log-Log Scale)

### Meta

- closes #338


bolt-contracts/src/contracts/BoltValidatorsV2.sol
```diff
@@ -0,0 +1,383 @@
...
+    /// used to identify entities with a 32-bit index instead of their full address.
+    mapping(uint32 => address) internal controllerIndexToAddress;
+    mapping(address => uint32) internal controllerAddressToIndex;
+    uint32 internal nextControllerIndex;
+
+    /// @notice Mapping of authorized operator index to its address. An authorized operator map is
+    /// used to identify entities with a 32-bit index instead of their full address.
+    mapping(uint32 => address) internal authorizedOperatorIndexToAddress;
+    mapping(address => uint32) internal authorizedOperatorToIndex;
+    uint32 internal nextAuthorizedOperatorIndex;
```
merklefruit: this is bad, planning to clean this tomorrow

mempirate: Note that any new storage variables MUST be added to the end of the already existing list of storage variables btw

bolt-contracts/script/holesky/admin/Upgrade.s.sol
```diff
@@ -103,6 +106,31 @@ contract UpgradeBolt is Script {
         console.log("BoltSymbioticMiddleware proxy upgraded from %s to %s", opts.referenceContract, upgradeTo);
     }
 
+    function upgradeBoltValidators() public {
+        address admin = msg.sender;
+        console.log("Upgrading BoltValidators with admin", admin);
+        // TODO: Validate upgrades with Upgrades.validateUpgrade
```
thedevbirb, line 113: Do you want to track this with an issue?

bolt-contracts/src/contracts/BoltManagerV2.sol
thedevbirb, line 1: Commenting it here but applies a bit everywhere: wdyt of having a type `CompressedHash` (or similar name) alias to `bytes20`? I found it really counter-intuitive to read `bytes20 pubkeyHash` etc

bolt-contracts/src/contracts/BoltValidatorsV2.sol
```diff
@@ -0,0 +1,295 @@
...
+        uint32 authorizedOperatorIndex = VALIDATORS.getOrInsertAuthorizedOperator(authorizedOperator);
+        uint32 controllerIndex = VALIDATORS.getOrInsertController(msg.sender);
+        uint256 pubkeysLength = pubkeyHashes.length;
+
+        for (uint32 i; i < pubkeysLength; i++) {
+            bytes20 pubkeyHash = pubkeyHashes[i];
+
+            if (pubkeyHash == bytes20(0)) {
+                revert InvalidPubkey();
+            }
```
thedevbirb, line 263: Is there any particular edge case happening if you save a compressed public key hash of zeros? I have two observations:
- even if there's a low probability, even with a genuine validator public key you could get a compressed hash consisting of only zero
- it is in the hot path consuming gas

bolt-contracts/src/contracts/BoltValidatorsV2.sol
```diff
@@ -0,0 +1,295 @@
...
+
+            VALIDATORS.insert(pubkeyHash, maxCommittedGasLimit, controllerIndex, authorizedOperatorIndex);
+            emit ValidatorRegistered(pubkeyHash);
```
thedevbirb, line 266: Is it really necessary to emit an event in every iteration? Since this unsafe registration the public keys might well be invalid. The client could link the public keys directly in a URI and we emit an event for that

thedevbirb (commented):
First review, will continue tmr!

bolt-contracts/script/holesky/admin/Upgrade.s.sol
merklefruit: This is blocked by the plugin not working due to some weird JS error loaded by the openzeppelin package.
We get around this by manually inspecting storage with Foundry which is not bad imo

bolt-contracts/src/contracts/BoltValidatorsV2.sol
mempirate: The space of possibilities is `2 ** 160` which I think is good enough for us

thedevbirb: As discussed internally we also rely on the `pubkeyHash` not being set to zero to tell if a record exists or not.

Also as outlined in a gas report of removing the event emission the gas saved is negligible compared to the call + storage write so I'm resolving this comment.

bolt-contracts/src/contracts/BoltValidatorsV2.sol
thedevbirb: Resolving this as gas saving are negligible compared to call + storage write.

bolt-contracts/src/interfaces/IBoltManagerV2.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+interface IBoltManagerV2 {
+    error InvalidQuery();
+    error OperatorAlreadyRegistered();
+    error OperatorNotRegistered();
+    error UnauthorizedMiddleware();
+    error InactiveOperator();
```
thedevbirb, line 10: By grepping over the codebase I've seen `InactiveOperator` is unused.

thedevbirb (approved):

bolt-contracts/src/interfaces/IBoltManagerV2.sol
mempirate: Damn missed this - added a TODO comment

PR #339, opened 2024-10-29
feat(sidecar): gross tip revenue metric

### LINKED ISSUE #245: Sidecar metrics for revenue
## Context
Proposers should have a way to track their revenue (in priority fees) from the commitments they make. 
Note that this should only be incremented when a preconf is actually included in their block.

The calculation should be the *effective* tip per gas x the gas the transaction actually consumed. Effective tip can be calculated with `tx.effective_tip_per_gas(basefee)` which requires the basefee of the block.

mempirate:
Closes #245

### DIFF
--- a/.github/workflows/bolt_sidecar_ci.yml
+++ b/.github/workflows/bolt_sidecar_ci.yml
@@ -2,6 +2,9 @@ name: Bolt Sidecar CI
 
 on:
   push:
+    branches:
+      - unstable
+      - main
     paths:
       - "bolt-sidecar/**"
   pull_request:
@@ -18,7 +21,7 @@ concurrency:
 jobs:
   cargo-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     env:
       RUST_BACKTRACE: 1
 
--- a/bolt-sidecar/src/telemetry/metrics.rs
+++ b/bolt-sidecar/src/telemetry/metrics.rs
@@ -20,6 +20,10 @@ const INCLUSION_COMMITMENTS_ACCEPTED: &str = "bolt_sidecar_inclusion_commitments
 const TRANSACTIONS_PRECONFIRMED: &str = "bolt_sidecar_transactions_preconfirmed";
 /// Counter for the number of validation errors; to spot most the most common ones
 const VALIDATION_ERRORS: &str = "bolt_sidecar_validation_errors";
+/// Counter that tracks the gross tip revenue. Effective tip per gas * gas used.
+/// We call it "gross" because in the case of PBS, it doesn't mean the proposer will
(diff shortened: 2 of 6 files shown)


.github/workflows/bolt_sidecar_ci.yml
```diff
@@ -2,6 +2,9 @@ name: Bolt Sidecar CI
 
 on:
   push:
+    branches:
+      - unstable
+      - main
```
thedevbirb, line 7: The `main` branch is unused. Any specific reason for adding it?

bolt-sidecar/src/telemetry/metrics.rs
```diff
@@ -81,6 +86,25 @@ impl ApiMetrics {
...
+            }
+
+            parts.push(tip as u64);
+
+            for part in parts {
+                counter!(GROSS_TIP_REVENUE).increment(part);
+            }
+        } else {
+            counter!(GROSS_TIP_REVENUE).increment(tip as u64);
+        }
```
thedevbirb, line 105: In my opinion this is a bit simpler/succint and doesn't distinguish between edge cases and regular cases which is nice.
```suggestion
         // Takes also into account tips larger than a u128 by splitting them into u64 parts
        let mut parts = Vec::new();
        let u64_max = u64::MAX as u128;

        while tip > 0 {
            // We can only add a max tip of u64::MAX.
            let min = tip.min(u64_max) as u64;
            parts.push(min);
            // Subtract the part we just added, if something is left then tip is going to be
            // greater than zero.
            tip = tip.saturating_sub(u64_max)
        }

        parts.push(tip as u64);

        for part in parts {
            counter!(GROSS_TIP_REVENUE).increment(part);
        }
```

thedevbirb (commented):
I've not seen the metric added anywhere in the sidecar code, is this PR still draft?

.github/workflows/bolt_sidecar_ci.yml
mempirate: We might use it in the future

bolt-sidecar/src/telemetry/metrics.rs
mempirate: True, but this always does an allocation

mempirate:
> I've not seen the metric added anywhere in the sidecar code, is this PR still draft?

Yes mb, converted

bolt-sidecar/src/telemetry/metrics.rs
thedevbirb: That's fair, also because with an u64 you can express a tip up to 18.44 ETH :D

merklefruit (approved):

thedevbirb (approved):
Great!

PR #340, opened 2024-10-29
Integrate with the Bolt registry to check if pubkeys are registered

### LINKED ISSUE #332: Integrate with the Bolt registry to check if pubkeys are registered
Currently we don't check whether the configured validator and operator public keys are registered in the Bolt sidecar.

We can do that at start time, by throwing an error if that's not the case.

thedevbirb:
Closes #332.

Introduces new checks for both the commitment signer key and the validator public keys which should correspond to an active operator and validator respectively in the `BoltManager` contract.

The check is performed at startup, when creating the `SidecarDriver::from_components`. This is a good spot because both signers (ECDSA, BLS) are abstracted via traits so the check can be safely done in a single pace.

 

## Summary by CodeRabbit

- **New Features**
	- Introduced `public_key` method in the `SignerECDSA` trait and its implementations, allowing retrieval of the public key as an address.
	- Added `BoltManager` for enhanced key verification in the `SidecarDriver`.
	- Launched the `onchain_registry` module, expanding on-chain management capabilities.
	- Implemented various utility methods in `BoltManager` for address retrieval and operator/validator verification.

- **Bug Fixes**
	- Improved readability of conditional statements and return formats in several methods.

- **Documentation**
	- Enhanced comments for clarity in the `make_path` and `test_keystore_signer` functions.

### DIFF
--- a/bolt-sidecar/src/chain_io/manager.rs
+++ b/bolt-sidecar/src/chain_io/manager.rs
@@ -26,6 +26,8 @@ pub struct BoltManager(BoltManagerContractInstance<Http<Client>, RootProvider<Ht
 impl BoltManager {
     /// Creates a new BoltRegistry instance. Returns `None` if a canonical BoltManager contract is
     /// not deployed on such chain.
+    ///
+    /// TODO: change after https://github.com/chainbound/bolt/issues/343 is completed
     pub fn from_chain<U: Into<Url>>(execution_client_url: U, chain: Chain) -> Option<Self> {
         let address = chain.manager_address()?;
         Some(Self::from_address(execution_client_url, address))
@@ -52,9 +54,11 @@ impl BoltManager {
     }
 
     /// Verify the provided validator public keys are registered in Bolt and are active
+    /// and their authorized operator is the given commitment signer public key
     pub async fn verify_validator_pubkeys(
         &self,
         keys: &[BlsPublicKey],
+        commitment_signer_pubkey: Address,
     ) -> eyre::Result<Vec<ProposerStatus>> {
         let hashes = utils::pubkey_hashes(keys);
 
@@ -69,6 +73,13 @@ impl BoltManager {
                             "validator with public key hash {:?} is not active in Bolt",
                             status.pubkeyHash
                         );
+                    } else if status.operator != commitment_signer_pubkey {
+                        bail!(
+                            "mismatch between commitment signer public key and authorized operator address for validator with public key hash {:?} in Bolt.\n - commitment signer public key: {:?}\n - authorized operator address: {:?}",
+                            status.pubkeyHash,
+                            commitment_signer_pubkey,
+                            status.operator
+                        );
                     }
                 }
 
@@ -78,7 +89,7 @@ impl BoltManager {
                 ContractError::TransportError(TransportError::ErrorResp(err)) => {
                     let data = err.data.unwrap_or_default();
                     let data = data.get().trim_matches('"');
-                    let data = Bytes::from_str(data).unwrap_or_default();
+                    let data = Bytes::from_str(data)?;
 
                     BoltManagerContractErrors::abi_decode(&data, true)?
                 }
--- a/bolt-sidecar/src/chain_io/manager.rs
+++ b/bolt-sidecar/src/chain_io/manager.rs
@@ -17,7 +17,7 @@ use BoltManagerContract::{BoltManagerContractErrors, BoltManagerContractInstance
 
 use crate::config::chain::Chain;
 
-use super::utils;
+use super::utils::{self, CompressedHash};
 
 /// A wrapper over a BoltManagerContract that exposes various utility methods.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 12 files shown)


bolt-sidecar/src/onchain_registry.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Returns the address of the canonical BoltManager contract for a given chain, if present
+    pub fn address(chain: Chain) -> Option<Address> {
+        match chain {
+            Chain::Holesky => Some(
+                Address::from_str("0x440202829b493F9FF43E730EB5e8379EEa3678CF")
+                    .expect("valid address"),
+            ),
+            _ => None,
+        }
+    }
```
merklefruit: why did you choose to place this here instead of in config::Chain where we already have all the chain specific stuff? To me it seems like a more appropriate place, wdyt?

mempirate: Agreed

bolt-sidecar/src/onchain_registry.rs
mempirate, line 1: Wdyt about putting this in its own `chain_io` module with a file per contract? I.e. manager & validators

bolt-sidecar/src/driver.rs
```diff
@@ -161,6 +161,16 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
...
         fetcher: C,
     ) -> eyre::Result<Self> {
+        // Verify the operator and validator keys with the bolt manager
+        if let Some(bolt_manager) =
+            BoltManager::from_chain(opts.execution_api_url.clone(), opts.chain.chain)
+        {
+            bolt_manager.verify_operator(commitment_signer.public_key()).await?;
+            bolt_manager
+                .verify_validator_pubkeys(&Vec::from_iter(constraint_signer.available_pubkeys()))
+                .await?;
```
mempirate, line 175: I think here we should also verify that the link between the validators and the operator is correct, i.e. the provided operator should be active for those validators

mempirate (changes requested):

bolt-sidecar/src/onchain_registry.rs
```diff
@@ -0,0 +1,163 @@
...
+        // the result should look like this:
+        //
+        // 0x00000000000000000000000000000000b427fd179b35ef085409e4a98fb3ab84ee29c689df5c64020eab0b20a4f91170f610177db172dc091682df627c9f4021
+        // |<---------- 16 bytes ---------->||<----------------------------------------- 48 bytes ----------------------------------------->|
+        onchain_pubkey_repr[16..].copy_from_slice(key);
+        onchain_pubkey_repr
+    }
+}
+
+sol! {
```
mempirate: In the future we should probably do something to keep these interfaces up to date, @coderabbitai please draft an issue

mempirate:
@coderabbitai review

bolt-sidecar/src/onchain_registry.rs
mempirate: Beast

bolt-sidecar/src/onchain_registry.rs
thedevbirb: Ok, I don't have a strong preference I first imagined this as an association Manager -> Chain rather than Chain -> Manager but the latter works well too

bolt-sidecar/src/driver.rs
thedevbirb: Right, great catch thank you

bolt-sidecar/src/chain_io/manager.rs
```diff
@@ -0,0 +1,169 @@
...
+    use ::hex::FromHex;
+    use alloy::{hex, primitives::Address};
+    use ethereum_consensus::primitives::BlsPublicKey;
+    use reqwest::Url;
+
+    use crate::config::chain::Chain;
+
+    use super::BoltManager;
+
+    #[tokio::test]
```
merklefruit, line 155: Can we add #[ignore] here? Otherwise CI won't pass

bolt-sidecar/src/chain_io/utils.rs
merklefruit, line 1: Very nice comments + tests, great job

merklefruit (approved):
Great!

bolt-sidecar/src/config/chain.rs
```diff
@@ -102,6 +104,17 @@ impl Chain {
             Chain::Kurtosis => [16, 0, 0, 56],
         }
     }
+
+    /// Returns the address of the canonical BoltManager contract for a given chain, if present
+    pub fn manager_address(&self) -> Option<Address> {
+        match self {
+            Chain::Holesky => Some(
+                Address::from_str("0x440202829b493F9FF43E730EB5e8379EEa3678CF")
```
mempirate: Can we put this in a constant / static?

mempirate (changes requested):
Last nit, great work

PR #353, opened 2024-11-04
chore(sidecar): warn in case of inconsistent gross revenue calculations

thedevbirb:


bolt-sidecar/src/state/execution.rs
```diff
@@ -464,9 +467,10 @@ impl<C: StateFetcher> ExecutionState<C> {
         if let Some(template) = self.remove_block_template(slot) {
             debug!(%slot, "Removed block template for slot");
             let hashes = template.transaction_hashes();
-            let receipts = self.client.get_receipts(&hashes).await?;
+            let receipts =
+                self.client.get_receipts(&hashes).await?.into_iter().flatten().collect::<Vec<_>>();
```
mempirate, line 471: I think this will allocate twice instead of once like previouy

thedevbirb: Good catch, fixed with 90891bfc1f4cdc544e43030893e045e91146d586

PR #355, opened 2024-11-04
chore: cleanup libraries, rm unused code

merklefruit:
* renamed `Config` to `BoltConfig` (keeps consistent with the pattern of FileName == ContractName)
* updated missing licenses and attribution comments where inaccurate
* deleted libraries that aren't used anymore


bolt-contracts/src/lib/bls/BLS12381.sol
thedevbirb, line 1: Can you add also whether this library has been audited or not? If so, add a URI in the comments. In this way it is easier to track for us which libraries might need audits on our side or not.

bolt-contracts/src/lib/bls/BLSSignatureVerifier.sol
thedevbirb, line 1: Same

bolt-contracts/src/lib/bls/BLS12381.sol
merklefruit: The issue is that I don't know this information. We can try asking or just wait

thedevbirb (approved):

PR #358, opened 2024-11-05
feat(cli): dirk delegations integration test

merklefruit:
This PR adds an integration test (`#[ignore]` by default) for generating and validating a delegation message with a local Dirk server. It has been implemented through a `test_util` module that could be handy in the future as well.


thedevbirb (approved):
Great!

PR #360, opened 2024-11-06
Drop `--validator-indexes` flag

thedevbirb:
It not needed anymore because we fetch the validator public key from the proposer duties, and then we check if we have the private key corresponding it (or delegated) to sign commitments with


merklefruit:
Duplicate of #361

PR #363, opened 2024-11-07
Add Dockerfile to `bolt` CLI

thedevbirb:


bolt-cli/Cargo.toml
```diff
@@ -40,3 +41,7 @@ tempfile = "3.13.0"
 
 [build-dependencies]
 tonic-build = "0.12.3"
+
+[[bin]]
+name = "bolt"
+path = "./src/main.rs"
```
merklefruit: wait why is this needed? the crate itself is just a binary we shouldn't need to specify this. What error did you get?

thedevbirb: This is the error I get when building the image without the `bin` array in the TOML:
```
...
=> ERROR [planner 2/2] RUN cargo chef prepare --recipe-path recipe.json                                                                                                                                                                       0.4s
------
 > [planner 2/2] RUN cargo chef prepare --recipe-path recipe.json:
0.382 Error: Failed to compute recipe
0.382
0.382 Caused by:
0.382     0: Cannot extract Cargo metadata
0.382     1: `cargo metadata` exited with an error: error: failed to parse manifest at `/app/Cargo.toml`
0.382
0.382        Caused by:
0.382          no targets specified in the manifest
0.382          either src/lib.rs, src/main.rs, a [lib] section, or [[bin]] section must be present
0.382
...
```
Which is weird considering there is a `src/main.rs` file.

PR #367, opened 2024-11-08
fix!(config): remove empty environment variables set like 'VAR=' when parsing options

thedevbirb:
It removes environment variables that are set as empty strings, i.e. like `MY_VAR=`. This useful to avoid unexpected edge cases and because we don't have options that make sense with an empty string value.


merklefruit (approved):
Very nice, and annoying that it's even a problem!

PR #370, opened 2024-11-08
chore(sidecar): cleanup + remove TODOs (continue)

merklefruit:
This PR addresses some small TODOs and cleans up the logic in the bolt manager pubkey verification.

### DIFF
--- a/bolt-sidecar/bin/sidecar.rs
+++ b/bolt-sidecar/bin/sidecar.rs
@@ -18,37 +18,22 @@ const BOLT: &str = r#"
 
 #[tokio::main]
 async fn main() -> eyre::Result<()> {
+    println!("{}", BOLT);
+
     read_env_file()?;
 
     let opts = Opts::parse();
 
     init_telemetry_stack(opts.telemetry.metrics_port())?;
 
-    println!("{BOLT}");
-
     info!(chain = opts.chain.name(), "Starting Bolt sidecar");
 
     if opts.constraint_signing.constraint_private_key.is_some() {
-        match SidecarDriver::with_local_signer(&opts).await {
-            Ok(driver) => driver.run_forever().await,
-            Err(err) => {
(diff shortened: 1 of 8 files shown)


bolt-sidecar/bin/sidecar.rs
```diff
@@ -18,37 +18,22 @@ const BOLT: &str = r#"
 
 #[tokio::main]
 async fn main() -> eyre::Result<()> {
+    println!("{}", BOLT);
```
thedevbirb, line 21: Curious -- do you prefer displaying the logo as the first thing or after all startup checks passed successfully?

I like the idea of seeing the logo knowing I've done everything well :D but maybe that's just me

bolt-sidecar/src/primitives/delegation.rs
```diff
@@ -57,15 +66,13 @@ impl SignableBLS for DelegationMessage {
 }
 
 /// read the delegations from disk if they exist and add them to the constraints client
-pub fn read_signed_delegations_from_file(
-    file_path: &PathBuf,
-) -> eyre::Result<Vec<SignedDelegation>> {
+pub fn read_signed_delegations_from_file(file_path: &PathBuf) -> Result<Vec<SignedDelegation>> {
```
thedevbirb: Can you keep `eyre::Result` as we discussed this morning?

thedevbirb (approved):

bolt-sidecar/bin/sidecar.rs
merklefruit: I get that, but I didn't like having other logs before the logo, such as:

[info] telemetry initialized!
LOGO
[...] other logs

instead it should be:
LOGO
[...] all logs

Not a blocker ofc, let's just pick one :)

thedevbirb: nvm, let's merge this! We can modify it at a later time if we wish

PR #377, opened 2024-11-12
Review logs on commitment deadline

thedevbirb:
The commitment deadline is triggered at every slot regardless of whether we're proposing or not, which was trigger a `WARN` log because the block template could not be found. This not unexpected behaviour, because block templates are created only when accepting commitments, and we do so only when we're the proposer for a block.

This PR reviews the logs in this section of the sidecar.

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -432,7 +432,6 @@ impl<C: StateFetcher> ExecutionState<C> {
 
     /// Commits the transaction to the target block. Initializes a new block template
     /// if one does not exist for said block number.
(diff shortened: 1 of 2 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -432,7 +432,6 @@ impl<C: StateFetcher> ExecutionState<C> {
 
     /// Commits the transaction to the target block. Initializes a new block template
     /// if one does not exist for said block number.
-    /// TODO: remove `pub` modifier once `try_commit` is fully implemented.
```
thedevbirb, line 435: This function `try_commit` does not exist, so I think this comment is stale.

merklefruit (approved):
LGTM

PR #378, opened 2024-11-12
chore(sidecar): bump alloy version to 0.6.3

merklefruit:
A node operator has encountered an edge case in Geth's `eth_feeHistory` endpoint. 
After some digging it turns out that this was a bug in Alloy's deserialization of the `FeeHistory` result.

It has already been fixed in Alloy last week here: https://github.com/alloy-rs/alloy/pull/1629
but we need to bump alloy in the sidecar in order to inherit it. 

This was due at some point, so here we go :)

Note: this also bumps the sidecar's toolchain to 1.82.0

### DIFF
--- a/bolt-sidecar/rust-toolchain.toml
+++ b/bolt-sidecar/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.81.0"
+channel = "1.82.0"
 profile = "default"
--- a/bolt-sidecar/src/builder/payload_builder.rs
+++ b/bolt-sidecar/src/builder/payload_builder.rs
@@ -1,24 +1,22 @@
 use std::fmt;
 
 use alloy::{
+    consensus::{Header, EMPTY_OMMER_ROOT_HASH},
     eips::{calc_excess_blob_gas, calc_next_block_base_fee, eip1559::BaseFeeParams},
-    primitives::{Address, Bytes, B256, U256},
-    rpc::types::{engine::ExecutionPayload as AlloyExecutionPayload, Block},
+    primitives::{Address, Bloom, Bytes, B256, B64, U256},
+    rpc::types::{engine::ExecutionPayload, Block, Withdrawal, Withdrawals},
 };
 use beacon_api_client::{BlockId, StateId};
 use hex::FromHex;
 use regex::Regex;
 use reqwest::Url;
-use reth_primitives::{
-    constants::BEACON_NONCE, proofs, BlockBody, Bloom, Header, SealedBlock, TransactionSigned,
-    Withdrawal, Withdrawals, EMPTY_OMMER_ROOT_HASH,
-};
+use reth_primitives::{proofs, BlockBody, SealedBlock, SealedHeader, TransactionSigned};
 use reth_rpc_layer::{secret_to_bearer_header, JwtSecret};
 use serde_json::Value;
 use tracing::trace;
 
 use super::{
-    compat::{to_alloy_execution_payload, to_reth_withdrawal},
+    compat::{to_alloy_execution_payload, to_alloy_withdrawal},
     BuilderError,
 };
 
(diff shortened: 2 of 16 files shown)


bolt-sidecar/src/builder/payload_builder.rs
```diff
@@ -124,8 +122,11 @@ impl FallbackPayloadBuilder {
         let prev_randao = self.get_prev_randao().await?;
         trace!(randao = ?prev_randao, "got prev_randao");
 
-        let parent_beacon_block_root =
-            self.beacon_api_client.get_beacon_block_root(BlockId::Head).await?;
+        let parent_beacon_block_root = B256::from_slice(
+            // TODO: compat: as_slice() from_slice() is necessary until we bump ethereum-consensus
+            // version to match alloy's.
+            self.beacon_api_client.get_beacon_block_root(BlockId::Head).await?.as_slice(),
```
merklefruit, line 128: Annoying

bolt-sidecar/src/crypto/ecdsa.rs
```diff
@@ -73,7 +73,10 @@ impl SignerECDSA for PrivateKeySigner {
     }
 
     async fn sign_hash(&self, hash: &[u8; 32]) -> eyre::Result<AlloySignature> {
-        Ok(Signer::sign_hash(self, hash.into()).await?)
+        let sig = Signer::sign_hash(self, hash.into()).await?;
+        // TODO: compat: this is necessary since alloy PrimitiveSignature and Signature
+        // are different types in the new version
+        Ok(AlloySignature::try_from(sig.as_bytes().as_ref()).unwrap())
```
merklefruit: Annoying

bolt-sidecar/src/signer/commit_boost.rs
```diff
@@ -122,9 +123,10 @@ impl CommitBoostSigner {
     /// Sign an object root with the Commit Boost domain.
     pub async fn sign_commit_boost_root(&self, data: [u8; 32]) -> SignerResult<BlsSignature> {
         // convert the pubkey from ethereum_consensus to commit-boost format
-        let pubkey = cb_common::signer::BlsPublicKey::from(
-            alloy::rpc::types::beacon::BlsPublicKey::from_slice(self.pubkey().as_ref()),
-        );
+        // TODO: compat: this parsing is the only way to obtain a Blspubkey. There are no other
+        // possible ways to obtain it without bumping Alloy version
+        let pubkey =
+            cb_common::signer::BlsPublicKey::from_ssz_bytes(self.pubkey().as_ref()).unwrap();
```
merklefruit: Extremely annoying and untested, I am not 100% sure this doesn't do something wrong.
Unfortunately there are no other ways to create this commit-boost type because `FixedBytes` from two different alloy versions aren't compatible between each other.

merklefruit (commented):
I left a few TODOs that can be solved when other dependencies eventually also update their alloy version.

bolt-sidecar/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.81.0"
+channel = "1.82.0"
```
thedevbirb, line 2: Can you update the toolchain also for the Dockerfile?

bolt-sidecar/src/crypto/ecdsa.rs
thedevbirb: Even if temporary can we avoid `unwrap` and use `expect` for production code?

bolt-sidecar/src/signer/commit_boost.rs
thedevbirb: As I've written above, can we use `expect` instead of `unwrap`? 

I think writing a quick unit test should clear your doubts, can you do that?

thedevbirb (changes requested):
Thanks for your service sir 🫡

thedevbirb (approved):

PR #382, opened 2024-11-13
feat(sidecar): --unsafe-disable-onchain-checks flag

thedevbirb:
For testing purposes, a flag to remove onchain checks of validators and operators with the `BoltManager` contract is introduced.


merklefruit (approved):
LGTM

PR #384, opened 2024-11-14
chore: cleanup monorepo outdated scripts

merklefruit:
* updated the justfile to be more in line with current devnet capabilities and useful commands
* removed outdated scripts that we don't use anymore
* removed dangling references to `demo` folder which was already removed


justfile
```diff
@@ -43,17 +31,26 @@ _restart-sidecar:
 inspect:
 	kurtosis enclave inspect bolt-devnet
 
+# format a rust crate. Must be run from the root of the crate's cargo directory
+fmt crate:
+  rustup toolchain install nightly-2024-10-03 > /dev/null 2>&1 && \
+  cd $(git rev-parse --show-toplevel)/{{crate}} && \
+  cargo +nightly-2024-10-03 fmt
+
+[private]
```
thedevbirb, line 35: Nice

justfile
```diff
@@ -149,31 +146,16 @@ build-images:
 	@just _build-bolt-boost
 
 # build the docker image for the bolt sidecar
+[private]
 _build-sidecar:
```
thedevbirb: According to the docs https://just.systems/man/en/private-recipes.html the attribute and the name with a leading underscore are equivalent. Can you choose one?

scripts/kurtosis_config.yaml
```diff
@@ -1,6 +1,6 @@
 network_params:
   seconds_per_slot: 2 # 2 seconds are the minimum for testing
-  genesis_delay: 0 # Apparently it seems needed to avoid timestamp issues on Helix
+  genesis_delay: 0 # This is needed to avoid timestamp issues on Helix
```
thedevbirb, line 3: We should re-check whether that's still the case but ok for now.

thedevbirb (approved):

PR #387, opened 2024-11-14
feat(sidecar): metadata endpoint

### LINKED ISSUE #247: Additional metadata endpoint
## Context
The sidecar is configured with various parameters like gas limit, min priority fee etc. We should expose this as an endpoint so that consumers like the RPC can be made aware of this.

thedevbirb:
Exposes an endpoint returning the sidecar operating limits in a `metadata` endpoint.

Closes #247


merklefruit (approved):
Looks great

PR #389, opened 2024-11-15
feat: add fast multi-platform build

merklefruit:
This PR adds support for building cross-platform binaries for `x86_64` and `aarch64` (`unknown-linux-gnu`).
We use cargo cross to produce a separate release binary for each architecture and successively load it
into a docker image directly.

Benefits:

* builds are fast, the part that takes longer is the linking at the end.
* no Docker overhead for build contexts, meaning also no need for janky docker cache layers
* lightweight output images, because we only ship the binary inside them

Downsides:

* since the images only contain the binary, it's harder to verify that they are running the correct binary. This is not a big issue because anyone wanting to verify their image integrity can build it locally from source.

## Usage

Available packages that can be built from this repo (so far): `bolt-sidecar`, `bolt-boost`.

Example to produce a cross-platform image with tag `unstable`:

```shell
just build-and-push-image bolt-sidecar unstable
```

Example to produce cross-platform images for all available packages mentioned above with tag `unstable`:

```shell
just build-and-push-all-images unstable
```

## Requisites to build images

* `cargo`
* `cross`
* `docker buildx`


thedevbirb (approved):
Awesome, super happy we got it working with the binary-based approach

merklefruit:
Converted to draft because there is an `openssl` issue when building bolt-boost for aarch64

bolt-boost/Cargo.toml
```diff
@@ -15,9 +18,7 @@ thiserror = "1.0.63"
 reqwest = "0.12.7"
 
 # crypto
-ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs", rev = "ec3073e", features = [
-    "sha2-asm",
-] }
+ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs", rev = "ec3073e" }
```
merklefruit, line 22: I had to remove the ssz-rs `sha2-asm` feature as it was breaking for cross-compiled `aarch64`.
There might be a way to do it but for now I had to disable it

merklefruit:
Images tagged `v0.3.0-alpha` have been produced from the latest commit.

thedevbirb (approved):
🫡

PR #398, opened 2024-11-19
chore: add supported holesky relay list

merklefruit:
Added Holesky relay list

### DIFF
--- a/testnets/holesky/README.md
+++ b/testnets/holesky/README.md
@@ -6,30 +6,30 @@ This document provides instructions for running the Bolt sidecar on the Holesky 
 
 <!-- vim-markdown-toc GFM -->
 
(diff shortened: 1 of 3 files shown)


testnets/holesky/README.md
```diff
@@ -685,6 +685,18 @@ docker compose -f telemetry.compose.yml down
 
 # Reference
 
+## Supported Relays
+
+Here is a list of Relays that support the Bolt constraints API and can be used
+as PBS relays when running the Bolt sidecar:
```
mempirate: Bolt mev boost instead of sidecar (or just bolt)

merklefruit:
Added relay URLs to linkspector ignore paths since we won't need to check them

PR #400, opened 2024-11-19
feat(bolt-cli): `validators` and `operators` commands

### LINKED ISSUE #392: feat(cli): Easily register validators (replace foundry scripts + better configuration)
This should be implemented as a new command under `bolt-cli/src/commands/`. 

The task consists of essentially porting [this script](https://github.com/chainbound/bolt/blob/unstable/bolt-contracts/script/holesky/validators/RegisterValidators.s.sol).

The first step is to add an interface for `BoltValidatorsV2` (just like the one we have for `BoltManager`).
In terms of the actual registration, we need to first read pubkeys from a json file and then call `batchRegisterValidatorsUnsafe`.

Note that there is already a CLI script called `pubkeys` that outputs a file that can be used as input for this.
It should also be necessary to hash the pubkeys to 20 bytes but we already have a rust implementation of that so it should be straight forward.

thedevbirb:
Closes #392


bolt-cli/src/cli.rs
```diff
@@ -131,6 +138,29 @@ pub struct SendCommand {
     pub devnet_sidecar_url: Option<Url>,
 }
 
+/// Command for registering validators in the Bolt network.
+#[derive(Debug, Clone, Parser)]
+pub struct RegisterCommand {
+    /// The chain for which the delegation message is intended.
+    #[clap(long, env = "CHAIN", default_value = "holesky")]
+    pub chain: Chain,
```
mempirate: Could we derive the chain from the RPC using eth_chainId? Cast uses the same pattern and it's a bit less verbose!

bolt-cli/src/cli.rs
```diff
@@ -131,6 +138,29 @@ pub struct SendCommand {
...
+
+    #[clap(long, env = "MAX_COMMITTED_GAS_LIMIT", default_value_t = 10_000_000)]
+    pub max_committed_gas_limit: u32,
```
mempirate: This should be required, let's remove the default value

mempirate (changes requested):
First review, following up tmr

bolt-cli/src/commands/register.rs
```diff
@@ -0,0 +1,127 @@
...
+enum BoltContract {
+    Validators,
+}
+
+impl RegisterCommand {
+    /// Run the `delegate` command.
+    pub async fn run(self) -> eyre::Result<()> {
+        let bolt_validators_address = bolt_validators_address(self.chain);
+
+        let pubkeys_file = std::fs::File::open(&self.pubkeys_path)?;
```
merklefruit: can you import File instead of qualifying?

bolt-cli/src/cli.rs
```diff
@@ -131,6 +138,29 @@ pub struct SendCommand {
...
+
+    #[clap(long, env = "AUTHORIZED_OPERATOR")]
+    pub authorized_operator: Address,
+
+    #[clap(long, env = "PUBKEYS_PATH", default_value = "pubkeys.json")]
+    pub pubkeys_path: PathBuf,
+
+    #[clap(long, env = "ADMIN_PRIVATE_KEY")]
+    pub admin_private_key: String,
```
merklefruit: I recommend actually parsing this as `B256` since it should automatically work from a hex string

bolt-cli/src/commands/register.rs
```diff
@@ -0,0 +1,127 @@
...
+        let keys: Vec<BlsPublicKey> = serde_json::from_reader(pubkeys_file)?;
+        let pubkey_hashes: Vec<_> = keys.iter().map(compress_bls_pubkey).collect();
+
+        let wallet: PrivateKeySigner =
+            self.admin_private_key.parse().wrap_err("invalid private key")?;
```
merklefruit:
```suggestion
        let wallet = PrivateKeySigner::from_bytes(self.admin_private_key).wrap_err("invalid private key")?;
```

bolt-cli/src/commands/register.rs
```diff
@@ -0,0 +1,127 @@
...
+        let transaction_signer = EthereumWallet::from(wallet);
+        let provider = ProviderBuilder::new()
+            .with_recommended_fillers()
+            .wallet(transaction_signer)
```
merklefruit:
```suggestion
        let provider = ProviderBuilder::new()
            .with_recommended_fillers()
            .wallet(EthereumWallet::from(wallet))
```

bolt-cli/src/commands/register.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        let result = call.send().await?.watch().await?;
+
+        println!("transaction hash: {:?}", result);
+
+        Ok(())
+    }
+}
+
+// PERF: this should be done at compile time
```
merklefruit: we could use phf for this https://docs.rs/phf/latest/phf/ but I think it's way overkill and I would opt for not adding extra dependencies. what do you think?

bolt-cli/src/commands/register.rs
```diff
@@ -0,0 +1,127 @@
...
+    use alloy::{
+        primitives::{Address, U256},
+        providers::{ext::AnvilApi, Provider, ProviderBuilder},
+        signers::k256::ecdsa::SigningKey,
+    };
+
+    use crate::cli::RegisterCommand;
+
+    #[tokio::test]
+    async fn test_register_validators() {
```
merklefruit: this is amazing, great job

bolt-cli/src/common/hash.rs
```diff
@@ -0,0 +1,27 @@
+// NOTE: Primitive types should be flattened into a single module
+// for cleaner imports and better organization.
```
merklefruit: rm

bolt-cli/src/common/mod.rs
```diff
@@ -16,6 +16,8 @@ pub mod keystore;
 /// Utilities for signing and verifying messages.
 pub mod signing;
 
+pub mod hash;
```
merklefruit:
```suggestion

/// Utilities for hashing messages and custom types
pub mod hash;
```

merklefruit (approved):
Left some nits, great overall!

bolt-cli/src/cli.rs
```diff
@@ -131,6 +138,29 @@ pub struct SendCommand {
...
 }
 
+/// Command for registering validators in the Bolt network.
+#[derive(Debug, Clone, Parser)]
+pub struct RegisterCommand {
+    /// The chain for which the delegation message is intended.
+    #[clap(long, env = "CHAIN", default_value = "holesky")]
+    pub chain: Chain,
+
+    #[clap(long, env = "MAX_COMMITTED_GAS_LIMIT", default_value_t = 10_000_000)]
```
merklefruit: I forgot to mention this but can you add doc comments to all the flags in this `RegisterCommand`?

bolt-cli/src/commands/register.rs
thedevbirb: Opting for a hashmap + struct approach.

mempirate (approved):
LGTM!

PR #402, opened 2024-11-20
chore(sidecar): mog unused dependencies with cargo-machete

merklefruit:
Used [cargo-machete](https://github.com/bnjbvr/cargo-machete) to remove unused dependencies in bolt-sidecar

```text
~/bolt/bolt-sidecar ❯ cargo machete
Analyzing dependencies of crates in this directory...
cargo-machete didn't find any unused dependencies in this directory. Good job!
```

I also took the opportunity to: 
* pin leftover unpinned git dependencies
* slightly refactor the sidecar entry binary
* remove `reth-rpc-layer` dependency by copying a 10-line utility function that uses all alloy types

### DIFF
--- a/bolt-sidecar/Cargo.toml
+++ b/bolt-sidecar/Cargo.toml
@@ -15,7 +15,6 @@ tokio = { version = "1", features = ["full"] }
 axum = { version = "0.7", features = ["macros"] }
 tower-http = { version = "0.5.2", features = ["timeout"] }
 axum-extra = "0.9.3"
-warp = "0.3.7"
 futures = "0.3"
 tokio-retry = "0.3.0"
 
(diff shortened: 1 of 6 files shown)


bolt-sidecar/Cargo.toml
```diff
@@ -24,21 +23,20 @@ blst = "0.3.12"
...
 secp256k1 = { version = "0.29.0", features = ["rand"] }
-ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs" }
+ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs", rev = "ec3073e" }
 ethereum_ssz = "0.5"
-ethereum_ssz_derive = "0.5"
 
 # alloy
-alloy = { version = "0.6.3", features = [
+alloy = { version = "0.6.4", features = [
   "full",
```
mempirate, line 31: Wonder if we need `full` here

merklefruit: if we remove it we'll have to add a bunch of other features in its place, but I haven't tried

mempirate:
Symbiotic links are failing - ignored them in #407 so can just rebase

PR #404, opened 2024-11-20
feat(cli): symbiotic integration

mempirate:


bolt-cli/src/commands/operators.rs
```diff
@@ -140,22 +186,124 @@ impl OperatorsCommand {
...
+                    }
+
+                    Ok(())
                 }
             },
         }
     }
 }
 
+fn request_confirmation() -> bool {
```
merklefruit: we can use inquire for this kind of thing. https://docs.rs/inquire/latest/inquire/struct.Confirm.html

thedevbirb: It's cool but considering current usage I'm okay with having a little function and not having another dependency

thedevbirb (approved):

bolt-cli/src/commands/operators.rs
merklefruit: Sounds good, however I'm OK adding this dependency down the line. It also enables interactivity in the CLI.

thedevbirb: Tracked in an issue!

PR #407, opened 2024-11-21
meta: add issue templates

mempirate:


merklefruit (approved):
I like this a lot! Let's give it a go

PR #418, opened 2024-11-25
feat: opt-in quick guide

merklefruit:
This PR adds a step-by-step quick guide to opt-in to bolt for node operators.

this version tries to address the shortcomings of the full guide in `holesky/README.md`
by skipping most explanations and being a bit more direct.


mempirate (approved):

thedevbirb (approved):
I like it and for me is good to go. We should remember that now this is another piece of docs we need to update.

PR #422, opened 2024-11-26
feat: `boltup` installer + script

merklefruit:
Quick installer for the `bolt` CLI tool.
Mostly copied from the amazing [foundryup](https://book.getfoundry.sh/getting-started/installation#using-foundryup) 👑.

> [!Note]
> the install script relies on a binary existing on the unstable branch, so this won't work until this PR is merged.

Try it with:

```sh
curl -L https://raw.githubusercontent.com/chainbound/bolt/boltup/boltup/install.sh | bash
```

I am not responsible if all your files are deleted, etc etc

### DIFF
--- a/boltup/boltup.sh
+++ b/boltup/boltup.sh
@@ -14,6 +14,9 @@ main() {
   need_cmd git
   need_cmd curl
 
+  # Loop as long as $1 (the first positional argument) is not empty, and match its 
+  # value against existing flags. If a flag is found, the argument is discarded 
+  # (via shift) and the value for the flag is read.
   while [[ -n $1 ]]; do
     case $1 in
       --)               shift; break;;
--- a/boltup/boltup.sh
+++ b/boltup/boltup.sh
@@ -1,0 +1,216 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# 'boltup' script for installing and managing the bolt CLI tool.
+#
+# This script is a heavily inspired by 'foundryup' (https://github.com/foundry-rs/foundry/blob/master/foundryup/foundryup)
+# since it is a great way to install CLI tools. Kudos to the Foundry team :)
+
+BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
+BOLT_DIR=${BOLT_DIR:-"$BASE_DIR/.bolt"}
+BOLT_BIN_DIR="$BOLT_DIR/bin"
+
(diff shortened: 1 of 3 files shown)


mempirate:
Nice, we'll host this on `https://boltprotocol.xyz/install.sh`

boltup/boltup.sh
```diff
@@ -0,0 +1,216 @@
...
+        exit 0
+        ;;
+      *)
+        warn "unknown option: $1"
+        usage
+        exit 1
+    esac; shift
+  done
+
+  CARGO_BUILD_ARGS=(--release)
```
mempirate: This is unused right?

boltup/boltup.sh
```diff
@@ -0,0 +1,216 @@
...
+# 'boltup' script for installing and managing the bolt CLI tool.
+#
+# This script is a heavily inspired by 'foundryup' (https://github.com/foundry-rs/foundry/blob/master/foundryup/foundryup)
+# since it is a great way to install CLI tools. Kudos to the Foundry team :)
+
+BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
+BOLT_DIR=${BOLT_DIR:-"$BASE_DIR/.bolt"}
+BOLT_BIN_DIR="$BOLT_DIR/bin"
+
+export RUSTFLAGS="${RUSTFLAGS:--C target-cpu=native}"
```
mempirate: Not used

mempirate (changes requested):
Nice, smol nits

boltup/boltup.sh
```diff
@@ -0,0 +1,212 @@
...
+      -h|--help)
+        usage
+        exit 0
+        ;;
+      *)
+        warn "unknown option: $1"
+        usage
+        exit 1
+    esac; shift
+  done
```
thedevbirb, line 36: Can this be moved to a separate, documented function?
By documentation I mean something like:

"This function loops as long as $1 (the first positional argument) is not empty, and matches its value against flags. If a flag is found, the argument is discarded (via `shift`) and the value for the flag is read."

I know, I should RTFM but having gently reminders of bash syntax is nice.

thedevbirb (approved):
wen nix

boltup/boltup.sh
merklefruit: I didn't move it but I added the comment, as I don't want to introduce untested behaviour in a script that is battle tested (this is what foundry has been using for 2+ years)

PR #424, opened 2024-11-26
fix: failing proof verification for blob transaction

### LINKED ISSUE #419: failing blob proof verification on devnet with bolt-boost
On the `holesky` branch proof verification is failing when sending a blob preconfirmation resuling in a locally built block. The error is given in bolt-boost, while the relay asserts the proof is correct.

```
2024-11-25T13:39:29.003893Z DEBUG get_header_with_proofs{slot=255}:handler{relay_id="helix_relay"}: bolt_boost::server: received new header latency=897.919µs block_hash=0xe19961234194d0a10710a110c36cd7d77b16fbd4e4f62b2ce0e83c3a166b9b2c value_eth="0.001284170277280550"
2024-11-25T13:39:29.004920Z ERROR get_header_with_proofs{slot=255}: bolt_boost::server: Failed to verify multiproof, skipping bid e=VerificationFailed relay_id="helix_relay"
```

thedevbirb:
Closes #419.

The issue was in the encoding logic of type 3 transactions to compute the proof data, regardless on whether they had a sidecar or not. 

In short, we were incorrectly discarding the signature fields before doing the EIP-2718 encoding to retrieve the raw bytes needed to compute the hash tree root of the transaction.
Computing an hash tree root with the wrong digest results in incorrectly computed leaves, as such proof verification was failing.

The PR adds also a test that runs in CI that verifies the process is carried out correctly, by loading test data correctly produced by the bolt-builder.


bolt-boost/src/proofs.rs
```diff
@@ -166,16 +120,95 @@ mod tests {
...
+        let inclusion_proof =
+            InclusionProofs { transaction_hashes, merkle_hashes: proof, generalized_indexes };
+
+        assert!(ssz_rs::multiproofs::verify_merkle_multiproof(
+            &leaves,
+            &inclusion_proof.merkle_hashes,
+            &inclusion_proof.generalized_indexes,
+            root
+        )
+        .is_ok());
```
merklefruit, line 181: beautiful, thanks for this

bolt-boost/src/types.rs
```diff
@@ -99,12 +99,7 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
         let transactions = value
             .transactions
             .iter()
-            .map(|tx| {
-                let envelope = TxEnvelope::decode_2718(&mut tx.as_ref())?;
-                let tx_hash_tree_root = calculate_tx_hash_tree_root(&envelope, tx)?;
-
-                Ok((*envelope.tx_hash(), tx_hash_tree_root))
-            })
+            .map(calculate_tx_proof_data)
```
merklefruit, line 102: clean

bolt-boost/src/types.rs
```diff
@@ -113,40 +108,47 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
 
 /// Calculate the SSZ hash tree root of a transaction, starting from its enveloped form.
```
merklefruit: is the comment here still accurate? seeing how the fn name changed.
I would add what is meant by "proof data"

bolt-boost/src/types.rs
```diff
@@ -113,40 +108,47 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
...
-            TxEip4844Variant::TxEip4844WithSidecar(tx) => {
-                use alloy_rlp::Encodable;
-                let mut out = Vec::new();
-                out.put_u8(0x03);
-                tx.tx.encode(&mut out);
+fn calculate_tx_proof_data(raw_tx: &Bytes) -> Result<(TxHash, HashTreeRoot), Eip2718Error> {
+    let is_type_3 = raw_tx
+        .first()
+        .ok_or(Eip2718Error::RlpError(alloy_rlp::Error::Custom("empty RLP bytes")))?
+        == &0x03;
```
merklefruit: maybe more readable, up to you
```suggestion
    let Some(is_type_3) = raw_tx.first().map(|type| type == &0x03) else {
    	return Eip2718Error::RlpError(alloy_rlp::Error::Custom("empty RLP bytes"));
    };
```

estensen (approved):

bolt-boost/src/types.rs
```diff
@@ -113,40 +108,47 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
...
+    };
+    let (tx, signature, tx_hash) = signed_tx.into_parts();
+    match tx {
+        TxEip4844Variant::TxEip4844(_) => {
+            // We have the type 3 variant without sidecar, we can safely hash tree root the raw
+            // RLP.
+            Ok((tx_hash, hash_tree_root_raw_tx(raw_tx.to_vec())))
+        }
+        TxEip4844Variant::TxEip4844WithSidecar(TxEip4844WithSidecar { tx, .. }) => {
+            // We strip out the sidecar and hash tree root the transaction
```
merklefruit:
```suggestion
            // We strip out the sidecar and compute the hash tree root of the transaction
```

bolt-boost/src/types.rs
```diff
@@ -113,40 +108,47 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
...
-        >::from(raw_tx.to_vec()))),
+    let envelope = TxEnvelope::decode_2718(&mut raw_tx.as_ref())?;
+    let TxEnvelope::Eip4844(signed_tx) = envelope else {
+        unreachable!("we have already checked it is not a type 3 transaction")
+    };
+    let (tx, signature, tx_hash) = signed_tx.into_parts();
+    match tx {
+        TxEip4844Variant::TxEip4844(_) => {
+            // We have the type 3 variant without sidecar, we can safely hash tree root the raw
+            // RLP.
```
merklefruit:
```suggestion
            // We have the type 3 variant without sidecar, we can safely compute the hash 
            // tree root of the transaction from the raw RLP bytes
```

bolt-boost/src/types.rs
```diff
@@ -113,40 +108,47 @@ impl TryFrom<ConstraintsMessage> for ConstraintsWithProofData {
...
+            Ok((tx_hash, hash_tree_root_raw_tx(raw_tx.to_vec())))
+        }
+        TxEip4844Variant::TxEip4844WithSidecar(TxEip4844WithSidecar { tx, .. }) => {
+            // We strip out the sidecar and hash tree root the transaction
+            let signed = Signed::new_unchecked(tx, signature, tx_hash);
+            let new_envelope = TxEnvelope::from(signed);
+            let mut buf = Vec::new();
+            new_envelope.encode_2718(&mut buf);
```
merklefruit, line 140: There should be an `encoded_2718` method that already does this
https://docs.rs/alloy/latest/alloy/consensus/enum.TxEnvelope.html#method.encoded_2718

merklefruit (changes requested):
Left a few nits, overall this is great and code quality is improved by a lot.

bolt-boost/src/types.rs
thedevbirb: Thanks, good catch I haven't revisited documentation.

bolt-boost/src/types.rs
thedevbirb: I like it! Thanks!

bolt-boost/src/types.rs
thedevbirb: Good to know, I don't see it available tho so I think it requires bumping alloy which is something I'd like to do in another PR if that's okay

merklefruit: Ah yeah just noticed bolt-boost has a pretty old version of alloy! No problem then

merklefruit (approved):
lgtm

PR #426, opened 2024-11-26
feat: built `bolt-cli` binaries in CI in each new release

merklefruit:
This PR adds a github action automatically triggered upon each new release. (we can also trigger it manually though).
It builds `bolt-cli` binaries for all platforms, archives them and uploads them to the release assets.

This is a slightly simplified version of https://github.com/foundry-rs/foundry/blob/master/.github/workflows/release.yml

### DIFF
--- a/.github/workflows/bolt_cli_release_bins.yml
+++ b/.github/workflows/bolt_cli_release_bins.yml
@@ -1,0 +1,141 @@
+name: Build and push Bolt CLI release binaries
+
+on:
+  push:
+    tags:
+      - "v*" # Push when version tags are pushed (e.g. v0.2.4)
+  release:
+    types:
+      - created # Push when a new release is created
+  workflow_dispatch: # allows manual triggering of the workflow
+
+env:
+  CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


.github/workflows/bolt_cli_release_bins.yml
```diff
@@ -0,0 +1,141 @@
...
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install Cross
+        run: |
+          cargo install cross --force
+
+      #   - name: Install Protoc
+      #     uses: arduino/setup-protoc@v3
```
estensen, line 67: remove commented out steps or do you need them later?

merklefruit: They might be needed if the workflow fails, but I don't know if it will, as I need to merge this to test it. 
I'll remove it in a further PR if everything is ok!

estensen (approved):

thedevbirb (approved):
Nice, I like we're getting better and better CI

PR #427, opened 2024-11-26
chore(bolt-boost): add more lints

estensen:
Please protest if you disagree and I'll avoid adding until we have reached agreement 🤝

### DIFF
--- a/bolt-boost/Cargo.toml
+++ b/bolt-boost/Cargo.toml
@@ -57,5 +57,10 @@ ssz_compat = { version = "0.5", package = "ethereum_ssz" }
 types = { git = "https://github.com/sigp/lighthouse", tag = "v5.3.0" }
 
 [lints.clippy]
+explicit_iter_loop = "warn"
+if_not_else = "warn"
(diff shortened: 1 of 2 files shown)


mempirate (approved):
I like it. We should turn this into a template for Chainbound Rust projects along with other defaults (formatters, toolchains, CI). Edit: could we also add this for the bolt-sidecar

mempirate:
Could you also do a PR for bolt?

estensen:
> I like it. We should turn this into a template for Chainbound Rust projects along with other defaults (formatters, toolchains, CI). Edit: could we also add this for the bolt-sidecar

Not sure what the best-practice for Rust monorepos to share tooling like linting is. I'll just copy it for now to bolt-sidecar and bolt-cli.

thedevbirb (approved):
AFAIK all clippy lints try to enforce either readability, idiomatic Rust, performance and possible footguns. All good stuff. Feel free to add them! We can eventually consider removing them at a later stage if we find some annoying ones.

PR #436, opened 2024-11-27
feat(scripts): add just lint

estensen:
To more easily run the same lint as in CI

### DIFF
--- a/justfile
+++ b/justfile
@@ -2,6 +2,14 @@
 default:
   @just --list --unsorted
 
+lint:
+	@just clippy bolt-cli
+	@just clippy bolt-boost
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -2,6 +2,14 @@
 default:
   @just --list --unsorted
 
+lint:
+	@just clippy bolt-cli
+	@just clippy bolt-boost
+	@just clippy bolt-sidecar
+
+clippy package:
+	cd {{ package }} && cargo clippy --all-targets --all-features -- -D warnings
```
merklefruit: can you add short doc comments starting with `# ` on the line above each command? it's useful when seeing the list of available commands with `just`

PR #487, opened 2024-11-28
fix(sidecar): nonce validation

mempirate:


estensen (approved):

bolt-sidecar/src/common/transactions.rs
```diff
@@ -58,17 +58,19 @@ pub fn validate_transaction(
...
         ));
     }
 
-    if transaction.nonce() > account_state.transaction_count {
+    if transaction.nonce() > account_state.transaction_count + 1 {
         return Err(ValidationError::NonceTooHigh(
-            account_state.transaction_count,
+            // Expected = transaction_count + 1
+            account_state.transaction_count + 1,
             transaction.nonce(),
```
thedevbirb, line 74: Let's see an example with an empty account, which has nonce 0.
- if it sends a valid preconfirmation, then its nonce is 0, and the `account_state.transactions_count` is 0 as well. This would run into a `NonceTooLow` error, so the logic of the first if is wrong.
- if it sends an invalid preconfirmation with nonce 1, then 1 is not greater than `account_state.transaction_count` + 1, i.e. $$1 \not\gt 1$$, and we don't enter in the second `if` statement which is incorrect

PR #488, opened 2024-11-28
chore(bolt-sidecar): extract calculate_max_accounts

estensen:
First sweep of trying to understand this part of the codebase.

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -195,12 +195,7 @@ impl<C: StateFetcher> ExecutionState<C> {
             client.get_chain_id()
         )?;
 
-        // Calculate the number of account states that can be cached by diving the configured max
-        // size by the size of an account state and its key.
-        let num_accounts = limits
-            .max_account_states_size
-            .get()
-            .div_ceil(size_of::<AccountState>() + size_of::<Address>());
+        let num_accounts = calculate_max_accounts(&limits);
 
         Ok(Self {
             basefee,
@@ -584,8 +579,15 @@ pub struct StateUpdate {
     pub block_number: u64,
 }
 
+// Calculate the number of account states that can be cached by diving the configured max
+// size by the size of an account state and its key.
+fn calculate_max_accounts(limits: &LimitsOpts) -> usize {
+    limits.max_account_states_size.get().div_ceil(size_of::<AccountState>() + size_of::<Address>())
+}
+
 #[cfg(test)]
 mod tests {
+    use super::*;
     use crate::{
         builder::template::StateDiff, config::limits::DEFAULT_MAX_COMMITTED_GAS,
         signer::local::LocalSigner,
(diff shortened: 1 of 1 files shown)


merklefruit:
This makes sense, however there is no reason for this calculation not to be done directly in the `ScoreMap` since it is generic. We can add a function to the `ScoreMap` like this:

```rust
    /// Creates an empty `ScoreMap` with maximum size in KiB.
    ///
    /// The maximum size is calculated based on the size of the key and value types.
    /// See also [std::collections::HashMap::new].
    #[inline]
    pub fn with_max_mem_size(max_size_kib: usize) -> Self {
        let max_len = max_size_kib.div_ceil(size_of::<V>() + size_of::<K>());
        Self { map: HashMap::<K, (V, isize)>::new(), max_len }
    }
```

and this works since all generic types in rust by default implement the `Sized` auto-trait

thedevbirb:
I think it has been a wrong choice on my side adding that logic, and we should allow in the configuration setting only the maximum length of the hashmap, and provide with a comment a raw estimate of the expected size. That is because accurately calculating the memory usage of an hashmap isn't easy. It should take into account the following:
- key size, that is `size_of::<K>()`;
- value size, which is `size_of::<V>()`;
- the hash of the key, which is the architecture pointer size.

More than that, each entry is stored in a bucket which is a pointer.

This is a great diagram output by ChatGPT for an `HashMap<usize, String>`:
```
HashMap Structure (in heap):
┌─────────────────────────────┐
│ Capacity (buckets)          │    -> usize (8 bytes)
│ Number of entries (len)     │    -> usize (8 bytes)
│ Load factor (internal)      │    -> usize (8 bytes)
│ Pointer to bucket array     │    -> *mut Bucket (8 bytes)
└─────────────────────────────┘

Bucket Array (in heap, size = capacity * bucket_size):
┌──────────────────────────────────────┐
│ Bucket 0                             │
│ ┌──────────────┬──────────────┬────┐ │
│ │ Hash (u64)   │ Key (usize)  │ Ptr│ │ -> String 0 (heap)
│ └──────────────┴──────────────┴────┘ │
│                                      │
│ Bucket 1                             │
│ ┌──────────────┬──────────────┬────┐ │
│ │ Hash (u64)   │ Key (usize)  │ Ptr│ │ -> String 1 (heap)
│ └──────────────┴──────────────┴────┘ │
│                                      │
│ ...                                  │
└──────────────────────────────────────┘
```
The number of buckets that an hashmap decides to allocate depends of course on the implementation and on a load factor. From the `std` code I see that when a capacity is specified it makes sure to add extend it so that at least 1/8 of the entries can be unset, and then it extended to the next power of two. So to compute a proper `max_len` from the specified max size we have to take into account this logic.

Consider how non-trivial the calculation can be, and how little this feature improves the sidecar overall, I propose indeed to change the behaviour of the flag inside `config/limits.rs`, either in this PR on in another one (and close this).

merklefruit:
Yeah it makes sense to keep it in "len" rather than "memory size". At the end of the day it needs to be a rough approximation and I was even considering removing the ability to configure it, so this works for me

merklefruit:
@estensen @thedevbirb how do you want to proceed w this? IMO the proposal by Lore works well

estensen:
might look at this later, but gonna focus on mainnet readiness instead rn

PR #490, opened 2024-11-28
chore(bolt-sidecar): extract compute_diffs()

estensen:
`validate_request()` is very long and hard to follow, so breaking it up and adding tests

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -565,13 +565,15 @@ pub struct StateUpdate {
     pub block_number: u64,
 }
 
-// From previous preconfirmations requests retrieve
-// - the nonce difference from the account state.
-// - the balance difference from the account state.
-// - the highest slot number for which the user has requested a preconfirmation.
-//
-// If the templates do not exist, or this is the first request for this sender,
-// its diffs will be zero.
+/// Calculate aggregated diffs for an account, given some block templates.
+/// 
+/// From previous preconfirmations requests retrieve
+/// - the nonce difference from the account state.
+/// - the balance difference from the account state.
+/// - the highest slot number for which the user has requested a preconfirmation.
+///
+/// If the templates do not exist, or this is the first request for this sender,
+/// its diffs will be zero.
 fn compute_diffs(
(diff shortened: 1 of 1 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -584,8 +563,37 @@ pub struct StateUpdate {
     pub block_number: u64,
 }
 
+// From previous preconfirmations requests retrieve
+// - the nonce difference from the account state.
+// - the balance difference from the account state.
+// - the highest slot number for which the user has requested a preconfirmation.
+//
+// If the templates do not exist, or this is the first request for this sender,
+// its diffs will be zero.
```
merklefruit:
```suggestion
/// Calculate aggregated diffs for an account, given some block templates.
/// 
/// From previous preconfirmations requests retrieve
/// - the nonce difference from the account state.
/// - the balance difference from the account state.
/// - the highest slot number for which the user has requested a preconfirmation.
///
/// If the templates do not exist, or this is the first request for this sender,
/// its diffs will be zero.
```

bolt-sidecar/src/state/execution.rs
```diff
@@ -609,7 +617,39 @@ mod tests {
...
+
+        let block_template = BlockTemplate { state_diff, signed_constraints_list: vec![] };
+
+        block_templates.insert(10, block_template);
+
+        let (nonce_diff, balance_diff, highest_slot) = compute_diffs(&block_templates, &sender);
+
+        assert_eq!(nonce_diff, 1);
+        assert_eq!(balance_diff, U256::from(2));
+        assert_eq!(highest_slot, 10);
```
thedevbirb: Can you add comments for the steps, so that I can read the test without simulating the code in my head?

thedevbirb (commented):
I like the idea

bolt-sidecar/src/state/execution.rs
merklefruit: Can't run a RISCV VM in your head in 2024? ngmi

estensen: better now?

merklefruit (approved):

PR #508, opened 2024-12-02
feat(cli): get validator status

mempirate:


estensen (approved):
Tested locally?

bolt-cli/src/commands/validators.rs
```diff
@@ -74,6 +74,46 @@ impl ValidatorsCommand {
...
+
+                let pubkey_hashes: Vec<_> = bls_pubkeys.iter().map(compress_bls_pubkey).collect();
+
+                let bolt_validators = BoltValidators::new(registry, provider);
+
+                for (hash, pubkey) in pubkey_hashes.iter().zip(bls_pubkeys.iter()) {
+                    match bolt_validators.getValidatorByPubkeyHash(*hash).call().await.map(|v| v._0)
+                    {
+                        Ok(info) => {
+                            info!(%pubkey, operator = %info.authorizedOperator, controller = %info.controller, gas_limit = info.maxCommittedGasLimit, "Validator registered");
```
merklefruit, line 109: I would print results to stdout or to a file instead of tracing

bolt-cli/src/contracts/bolt.rs
```diff
@@ -4,13 +4,25 @@ sol! {
     #[allow(missing_docs)]
     #[sol(rpc)]
     interface BoltValidators {
+        struct ValidatorInfo {
```
merklefruit, line 9: you can add `#[derive(Debug, Serialize, Deserialize)]` and then encode it to JSON if necessary

merklefruit (changes requested):
Looks good, 2 nits

bolt-cli/src/commands/validators.rs
thedevbirb, line 1: Can you add a test for it built upon `test_register_validators`?

mempirate:
> Tested locally?

Yes sir!

bolt-cli/src/commands/validators.rs
mempirate: We've been printing all output with tracing, so for consistency let's keep this

merklefruit (approved):

PR #509, opened 2024-12-02
feat(cli): Dirk multi-node delegation support

merklefruit:
This PR adds support for creating signed delegations from multi-node Dirk instances (aka `DistributedAccount`).

Unfortunately, the BLS library we have used until now (`BLST`) doesn't support M-N threshold signature recovery.
See the related issue: https://github.com/supranational/blst/issues/123

Since this operation is not overly complex, I've decided to implement it manually in `recover_signature.rs`,
following the [C reference implementation in MCL](https://github.com/herumi/mcl/blob/328e26f45ba565d031f9570e68e3d61836a17d7c/include/mcl/lagrange.hpp#L16) directly for the Lagrange interpolation

The implementation is tested by generating a signed delegation against a 2-of-3 Dirk setup running locally and then verifying the signature against the combined public key.

One important caveat for testing is that the multi-node setup only works if you modify your `/etc/hosts` file to add aliases for localhost to `localhost-1`, `localhost-2`, `localhost-3` because it's the only way to have different valid TLS certificates on the same machine.


bolt-cli/src/commands/delegate/dirk.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Test generating signed delegations using a remote Dirk signer (multi-node instance).
+    /// This test requires multiple instances of Dirk to be running.
+    ///
+    /// ```shell
+    /// cargo test --package bolt --bin bolt -- commands::delegate::tests::test_delegation_dirk_multi
+    /// --exact --show-output --ignored --nocapture
+    /// ```
+    #[tokio::test]
+    #[ignore = "Requires Dirk to be installed on the system"]
+    async fn test_delegation_dirk_multi() -> eyre::Result<()> {
```
merklefruit, line 182: this is WIP.

bolt-cli/src/common/dirk/distributed.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        for participant in &self.participants {
+            if signatures_with_ids.len() >= self.threshold {
+                break;
+            }
+
+            let url = participant_url(participant);
+            let mut conn = Dirk::connect(url.clone(), self.credentials.clone()).await?;
+
+            // TODO: might have to unlock the account here
```
thedevbirb: Is this stale?

thedevbirb (approved):
Great job. BLS is so cool

bolt-cli/src/commands/delegate/dirk.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    /// Test generating signed delegations using a remote Dirk signer (single instance).
+    ///
+    /// ```shell
+    /// cargo test --package bolt --bin bolt -- commands::delegate::tests::test_delegation_dirk_single
+    /// --exact --show-output --ignored --nocapture
+    /// ```
+    #[tokio::test]
+    #[ignore = "Requires Dirk to be installed on the system"]
+    async fn test_delegation_dirk_single() -> eyre::Result<()> {
```
mempirate, line 145: Very nice tests

bolt-cli/src/common/dirk/distributed.rs
```diff
@@ -0,0 +1,98 @@
...
+
+            // Every remote signer must sign the same message
+            let signature = match conn.request_signature(account_name.clone(), hash, domain).await {
+                Ok(signature) => signature,
+                Err(err) => {
+                    warn!(?err, "Failed to sign message with remote signer at {}", url);
+                    continue;
+                }
+            };
```
mempirate: Can we do this concurrently by pushing them into a `FuturesUnordered` or something similar for each participant?

bolt-cli/src/common/dirk/test_util.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Returns the DIRK client URL and credentials, and the corresponding server process handle.
+pub async fn start_single_dirk_test_server() -> eyre::Result<(String, TlsCredentials, Child)> {
+    try_init_tls_provider();
+
+    // Check if dirk is installed (in $PATH)
+    if Command::new("dirk").arg("--help").status().is_err() {
+        bail!("DIRK is not installed in $PATH");
+    }
+
+    let test_data_dir = env!("CARGO_MANIFEST_DIR").to_string() + "/test_data/dirk_single";
```
mempirate: Nit but can use `concat!(env!("CARGO_MANIFEST_DIR"), "/test_data/dirk_single");`

mempirate (changes requested):

bolt-cli/src/common/dirk/distributed.rs
merklefruit: Yes, removing

bolt-cli/src/common/dirk/test_util.rs
merklefruit: nice, done

bolt-cli/src/common/dirk/distributed.rs
merklefruit: Yes but it would need to be throttled to avoid DDoSing dirk

merklefruit: done in the latest commit

mempirate (approved):
LGTM! Great work

PR #510, opened 2024-12-03
chore(holesky): pin lighthouse image, change docs

thedevbirb:
Minor changes to reflect the new Lighthouse release that removes the `always-prefer-builder-payload` flag. The image is now pinned, and docs have been changed a bit.

The section on what flags to provide is a bit generic but it's okay for now. Ideally we should add a comprehensive guide on which flags to enable on all beacon clients/validators clients.

### DIFF
--- a/testnets/holesky/README.md
+++ b/testnets/holesky/README.md
@@ -12,33 +12,33 @@ This document provides instructions for running Bolt on the Holesky testnet.
 
 <!-- vim-markdown-toc GFM -->
 
-- [Prerequisites](#prerequisites)
-- [On-Chain Registration](#on-chain-registration)
-  - [Validator Registration](#validator-registration)
-    - [Registration Steps](#registration-steps)
-  - [Bolt Network Entrypoint](#bolt-network-entrypoint)
-  - [Operator Registration](#operator-registration)
-    - [Symbiotic Registration Steps](#symbiotic-registration-steps)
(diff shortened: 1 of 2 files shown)


testnets/holesky/README.md
```diff
@@ -71,21 +71,15 @@ client implementations to download and run them.
...
->    --always-prefer-builder-payload
->    --builder-fallback-disable-checks
->    ```
->
-> In other clients like Vouch, the same can be achieved by setting the
-> `builder-boost-factor` to a large value like `18446744073709551615` (`2**64 -
-1`).
+> 2. the node and the validator client will always prefer the builder payload,
+>    and try to use the builder API. For instance, in Lighthouse this can be
+>    done by setting the `builder-fallback-disabled-checks` flag and the
```
mempirate, line 80: Where is this flag defined?

mempirate (approved):

testnets/holesky/README.md
mempirate: Typo: should be `builder-fallback-disable-checks`

mempirate (changes requested):

thedevbirb:
@mempirate I have to re-request a review otherwise I cannot merge it -- can we loosen up the requirements for merging for us core devs?

mempirate (approved):

PR #511, opened 2024-12-03
chore(sidecar): bump alloy version to 0.7.2

merklefruit:
Now that `PooledTransactionsElement` implements `alloy::consensus::Transaction` trait, we can remove some of the functions in our ad-hoc `TransactionExt` trait. However, we still need some of them, so we need to keep our trait as well.


mempirate (approved):
Lgtm! Can you redeploy this version of the sidecar on holesky?

merklefruit:
Sure, what docker tag should I use for it?

PR #512, opened 2024-12-03
feat(sidecar): fallback block builder refactor

merklefruit:
This PR contains a refactor to the bolt-sidecar `builder` module to generalize over the EL client used for fallback block building. The refactor also comes with:

* clients separated in their own `client` module: `execution`, `beacon`, `engine` and `constraints`
* more documentation on the block building process


thedevbirb (approved):
Nice refactor, looks good to me

bolt-sidecar/src/builder/fallback/engine_hinter.rs
```diff
@@ -0,0 +1,259 @@
...
+                exec_payload,
+                ctx.blob_versioned_hashes.clone(),
+                ctx.parent_beacon_block_root,
+            )
+            .await?;
+
+        let validation_error = match payload_status.status {
+            PayloadStatusEnum::Valid => return Ok(EngineApiHint::ValidPayload),
+            PayloadStatusEnum::Invalid { validation_error } => validation_error,
+            PayloadStatusEnum::Syncing | PayloadStatusEnum::Accepted => {
```
mempirate, line 115: Why is `Accepted` an error?

bolt-sidecar/src/builder/fallback/engine_hinter.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Populate the new hint value in the context.
+    pub fn populate_new(&mut self, hint: EngineApiHint) {
+        match hint {
+            EngineApiHint::ValidPayload => { /* No-op */ }
+
+            // If we receive a block hash hint, set it and keep it for the next one.
+            // This should not happen, but in case it does, it doesn't break the flow.
+            EngineApiHint::BlockHash(hash) => self.block_hash = Some(hash),
+
+            // For regular hint types, set the value and reset the block hash
```
mempirate, line 169: Why do we reset the block hash hint here every time?

mempirate (changes requested):
Great stuff, only minor comments

bolt-sidecar/src/builder/fallback/payload_builder.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        // Fetch the execution client info from the engine API in order to know what hint
+        // types the engine hinter can parse from the engine API responses.
+        let el_client_code = self.engine_hinter.engine_client_version().await?[0].code;
+        debug!(client = %el_client_code.client_name(), "Fetched execution client info");
+
+        // Fetch required head info from the beacon chain
+        let parent_beacon_block_root = self.beacon_api.get_parent_beacon_block_root().await?;
+        let withdrawals = self.beacon_api.get_expected_withdrawals_at_head().await?;
+        let prev_randao = self.beacon_api.get_prev_randao().await?;
```
mempirate: Let's do this concurrently, these calls are all independent

mempirate (changes requested):
Great job, just one req

bolt-sidecar/src/builder/fallback/engine_hinter.rs
merklefruit: Because we don't want the EL to accept this block. According to the engine specs: https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_newpayloadv1 the ACCEPTED code MUST be returned only if the following conditions are met:
* all transactions have non-zero length
* the blockHash of the payload is valid
* the payload doesn't extend the canonical chain
* the payload hasn't been fully validated
* ancestors of a payload are known and comprise a well-formed chain

in our case our block extends the canonical chain so it will be marked as VALID.

bolt-sidecar/src/builder/fallback/engine_hinter.rs
merklefruit: It really depends on how the EL validation is structured, but me and @thedevbirb found that sometimes Geth would return a mismatching hash error after populating some extra field (e.g. state root) which means we can't get more information out of it (as it kept returning `mismatching block hash` and no other hints). So when this happens we set the block hash to the one hinted by Geth and then proceed with the next hints. 

As the comment mentions, this shouldn't happen and it's more of an edge case

estensen (approved):

mempirate (approved):

PR #516, opened 2024-12-04
chore(docs): add vouch config paragraph

merklefruit:
This PR adds a small paragraph on Vouch configuration for validators.
Looking to add more details but this should be in the right direction.


mempirate (approved):
lgtm, please add entry in ToC

PR #521, opened 2024-12-06
chore: use more specific error types

merklefruit:
Small chore.

### DIFF
--- a/bolt-sidecar/src/api/commitments/spec.rs
+++ b/bolt-sidecar/src/api/commitments/spec.rs
@@ -34,6 +34,9 @@ pub enum CommitmentError {
     /// Duplicate request.
     #[error("Duplicate request")]
     Duplicate,
+    /// No available public key to sign commitment request with for a given slot.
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/api/commitments/spec.rs
```diff
@@ -34,6 +34,9 @@ pub enum CommitmentError {
     /// Duplicate request.
     #[error("Duplicate request")]
     Duplicate,
+    /// No available public key to sign commitment request with for a given slot.
+    #[error("No available public key to sign request with (slot {0})")]
+    NoAvailablePubkeyForSlot(u64),
```
mempirate: These errors shouldn't be propagated back to the user imo. It should probably be logged on the sidecar but return internal server error or something generic like that.

merklefruit: Got it!

thedevbirb (approved):

PR #537, opened 2024-12-09
chore(ci): run on self-hosted

estensen:
Should improve CI times. We could also consider breaking up CI and have separate jobs for clippy/lint and test.


merklefruit (approved):
let's see how it goes!

PR #540, opened 2024-12-09
chore(sidecar/holesky): add preset for delegation, env docs

thedevbirb:
This PR introduces a `sidecar-delegations-preset.env.example` file that, as the name suggests, offers an example environment variable file with only the delegations signing options enabled.
The quick guide README has been changed accordingly, along with the documentation of some of the environment variables available.

This is how the "Docker mode" section of the quick guide looks like now:

image

### DIFF
--- a/testnets/holesky/presets/sidecar-delegations-preset.env.example
+++ b/testnets/holesky/presets/sidecar-delegations-preset.env.example
@@ -25,9 +25,6 @@ BOLT_SIDECAR_FEE_RECIPIENT=
 # Secret ECDSA key used to sign commitment messages on behalf of your validators.
 # This MUST be set to the private key of your operator address registered in a restaking protocol.
 BOLT_SIDECAR_COMMITMENT_PRIVATE_KEY=
-# Secret BLS key to sign fallback payloads with. This can be any key. You can generate one
-# with the `bolt` CLI tool, using `bolt generate bls`.
-BOLT_SIDECAR_BUILDER_PRIVATE_KEY=
 
 # Commitments limits
 # Max number of commitments to accept per block
--- a/testnets/holesky/presets/sidecar-delegations-preset.env.example
+++ b/testnets/holesky/presets/sidecar-delegations-preset.env.example
@@ -19,11 +19,11 @@ BOLT_SIDECAR_CONSTRAINTS_API_URL="http://bolt-mev-boost-holesky:18551"
 # either be a hex-encoded string or a file path to a file containing the
 # hex-encoded secret.
 BOLT_SIDECAR_ENGINE_JWT_HEX=
-# The fee recipient address for fallback blocks
+# The fee recipient address for locally-built fallback blocks. It should be the same as the
+# one set for your validators.
 BOLT_SIDECAR_FEE_RECIPIENT=
-# Secret ECDSA key used to sign commitment messages on behalf of your validators. The public
-# key associated is an operator registered in the `BoltManager` contract and in a restaking
-# protocol.
+# Secret ECDSA key used to sign commitment messages on behalf of your validators.
+# This MUST be set to the private key of your operator address registered in a restaking protocol.
 BOLT_SIDECAR_COMMITMENT_PRIVATE_KEY=
 # Secret BLS key to sign fallback payloads with. This can be any key. You can generate one
 # with the `bolt` CLI tool, using `bolt generate bls`.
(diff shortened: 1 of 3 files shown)


testnets/holesky/presets/sidecar-delegations-preset.env.example
```diff
@@ -0,0 +1,60 @@
...
+# The port from which the Bolt sidecar will receive Builder-API requests from the Beacon client
+BOLT_SIDECAR_CONSTRAINTS_PROXY_PORT=18550
+# URL to forward the constraints produced by the Bolt sidecar to a server
+# supporting the Constraints API, such as an MEV-Boost fork
+BOLT_SIDECAR_CONSTRAINTS_API_URL="http://bolt-mev-boost-holesky:18551"
+# The JWT secret token to authenticate calls to the engine API. It can be
+# either be a hex-encoded string or a file path to a file containing the
+# hex-encoded secret.
+BOLT_SIDECAR_ENGINE_JWT_HEX=
+# The fee recipient address for fallback blocks
```
merklefruit:
```suggestion
# The fee recipient address for locally-built fallback blocks (should be the same as the one set for your validators)
```

testnets/holesky/presets/sidecar-delegations-preset.env.example
```diff
@@ -0,0 +1,60 @@
...
+BOLT_SIDECAR_FEE_RECIPIENT=
+# Secret ECDSA key used to sign commitment messages on behalf of your validators. The public
+# key associated is an operator registered in the `BoltManager` contract and in a restaking
+# protocol.
+BOLT_SIDECAR_COMMITMENT_PRIVATE_KEY=
```
merklefruit:
```suggestion
# Secret ECDSA key used to sign commitment messages on behalf of your validators.
# This MUST be set to the private key of your operator address registered in a restaking protocol.
BOLT_SIDECAR_COMMITMENT_PRIVATE_KEY=
```

testnets/holesky/presets/sidecar-delegations-preset.env.example
```diff
@@ -0,0 +1,60 @@
...
+# Secret BLS key to sign fallback payloads with. This can be any key. You can generate one
+# with the `bolt` CLI tool, using `bolt generate bls`.
+BOLT_SIDECAR_BUILDER_PRIVATE_KEY=
```
merklefruit: I am thinking that we should use a random one as default and not expose this variable in the preset, because in 99% of cases validators don't care about this and if they do it's OK for it to be an advanced feature IMO.

If you agree I can draft an issue

testnets/holesky/presets/sidecar-delegations-preset.env.example
```diff
@@ -0,0 +1,60 @@
...
+BOLT_SIDECAR_MAX_COMMITTED_GAS_PER_SLOT=10_000_000
+# Min priority fee to accept for a commitment
+BOLT_SIDECAR_MIN_PRIORITY_FEE=4_000_000_000 # 4 Gwei = 4 * 10^9 wei
+
+# Chain configuration
+# Chain on which the sidecar is running
+BOLT_SIDECAR_CHAIN="holesky"
+# The slot time duration in seconds. If provided, it overrides the default for
+# the selected [chain]
+BOLT_SIDECAR_SLOT_TIME=12
```
merklefruit: no reason to have this here in my opinion

merklefruit (changes requested):
I like it!

testnets/holesky/presets/sidecar-delegations-preset.env.example
mempirate: Agreed

testnets/holesky/presets/sidecar-delegations-preset.env.example
thedevbirb: I agree with the point you make. Why would you need to draft an issue for it?

merklefruit: because BOLT_SIDECAR_BUILDER_PRIVATE_KEY currently has no default value in the sidecar, we need to add it

merklefruit: it can be done here as we already have a Default impl, gtg

mempirate (approved):

merklefruit (approved):

PR #545, opened 2024-12-10
fix(holesky): entrypoint logic to not replace envs when empty

thedevbirb:
There was an error in the entrypoint logic that was replacing some environment variable names with the volume mount paths even when they were empty. That should only happen when they're set.
As such, the sidecar was throwing an error because it seeing incompatible flags provided.

### DIFF
--- a/bolt-sidecar/src/config/mod.rs
+++ b/bolt-sidecar/src/config/mod.rs
@@ -21,7 +21,6 @@ use telemetry::TelemetryOpts;
 /// Operating limits for commitments and constraints.
 pub mod limits;
 use limits::LimitsOpts;
-use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


bolt-sidecar/src/config/mod.rs
```diff
@@ -144,7 +143,7 @@ fn remove_empty_envs() -> eyre::Result<()> {
     for item in env::vars() {
         let (key, val) = item;
         if val.trim().is_empty() {
-            debug!("removing empty env var: {}", key);
+            println!("Removing empty environment variable with key: {}", key);
```
merklefruit: if this has been tested and works, can you remove this? I would try to avoid printing to stdout if possible

merklefruit (approved):

PR #546, opened 2024-12-10
chore(sidecar): add small test

merklefruit:


thedevbirb (approved):

merklefruit:
@mempirate could you install protoc on `self-hosted` to fix CI? 

required packages for ubuntu should be `protobuf-compiler libprotobuf-dev`

PR #550, opened 2024-12-11
feat(bolt-cli): log if operators have collateral

estensen:
Make it easier to debug and verify sufficient collateral.


bolt-cli/src/contracts/mod.rs
```diff
@@ -145,4 +156,12 @@ const HOLESKY_DEPLOYMENTS: Contracts = Contracts {
...
         },
     },
+    collateral: Token {
+        wst_eth: address!("8d09a4502Cc8Cf1547aD300E066060D043f6982D"),
+        r_eth: address!("7322c24752f79c05FFD1E2a6FCB97020C1C264F1"),
+        st_eth: address!("3F1c547b21f65e10480dE3ad8E19fAAC46C95034"),
+        w_eth: address!("94373a4919B3240D86eA41593D5eBa789FEF3848"),
+        cb_eth: address!("8720095Fa5739Ab051799211B146a2EEE4Dd8B37"),
+        m_eth: address!("e3C063B1BEe9de02eb28352b55D49D85514C67FF"),
+    },
```
merklefruit: I think that turning this into an array of tuples would allow for iteration, allowing us to avoid duplicating the function call for each new token.

something like:

```rust
collateral: [
	("wst_eth".to_string(), address!("8d09a4502Cc8Cf1547aD300E066060D043f6982D"),
         ...
]

// then:
for (name, address) in collateral {
	let stake = bolt_manager.getOperatorStake(operator, address).call().await?._0;
        // ...
}

```

estensen: Is enums the best way of working around this?

```rust
#[derive(Clone, Serialize, Deserialize, Debug)]
pub struct Token(pub Vec<(String, Address)>);

// allocations are not allowed in constants
collateral: Token(vec![
    ("wst_eth".to_string(), address!("8d09a4502Cc8Cf1547aD300E066060D043f6982D")),
```

```rust
#[derive(Clone, Serialize, Deserialize, Debug)]
// lifetime may not live long enough
// requires that `'de` must outlive `'static`
pub struct Token(pub [(&'static str, Address); 6]);

collateral: Token([
    ("wst_eth", address!("8d09a4502Cc8Cf1547aD300E066060D043f6982D")),
```

merklefruit: No, it's most likely Lazy statics; constant allocs in rust are very limited unfortunately

merklefruit:
```rust
// cargo.toml
lazy_static = "1.5.0"
```

```rust
pub struct Contracts {
    pub bolt: Bolt,
    pub symbiotic: Symbiotic,
    pub eigen_layer: EigenLayer,
    pub collateral: Vec<(String, Address)>,
}
```

```rust
lazy_static! {
	public static ref HOLESKY_DEPLOYMENTS: Contracts = Contracts {
		// here you can allocate more freely
    }
}
```

bolt-cli/src/commands/operators.rs
```diff
@@ -254,6 +254,34 @@ impl OperatorsCommand {
...
+                            total_collateral += stake;
+                            info!(?address, token = %name, amount = ?stake, "Operator has collateral");
+                        }
+                    }
+                    if total_collateral > Uint::from(0) {
+                        info!(?address, total_collateral=?total_collateral, "Total operator collateral");
+                    } else {
+                        warn!(?address, "Operator has no collateral");
+                    }
+
```
thedevbirb: This is good, I'd also write to stdout whether the operator is "active" i.e. its total collateral is greater than 1 ether.

bolt-cli/src/contracts/mod.rs
```diff
@@ -106,43 +108,53 @@ pub fn deployments_for_chain(chain: Chain) -> Contracts {
...
-        delegation_manager: address!("A44151489861Fe9e3055d95adC98FbD462B948e7"),
-        strategy_manager: address!("dfB5f6CE42aAA7830E94ECFCcAd411beF4d4D5b6"),
-        middleware: address!("a632a3e652110Bb2901D5cE390685E6a9838Ca04"),
-        supported_strategies: EigenLayerStrategies {
-            st_eth: address!("7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3"),
-            r_eth: address!("3A8fBdf9e77DFc25d09741f51d3E181b25d0c4E0"),
-            w_eth: address!("80528D6e9A2BAbFc766965E0E26d5aB08D9CFaF9"),
-            cb_eth: address!("70EB4D3c164a6B4A5f908D4FBb5a9cAfFb66bAB6"),
-            m_eth: address!("accc5A86732BE85b5012e8614AF237801636F8e5"),
+lazy_static! {
```
thedevbirb, line 111: What's the advantage of using `lazy_static` in this context?

estensen: see thread
https://github.com/chainbound/bolt/pull/550#discussion_r1880417436

bolt-cli/src/commands/operators.rs
estensen: ok to hardcode this or should >1ETH be retrieved from somewhere? Is it solidified?

merklefruit: for now ok to hardcode it imo

thedevbirb: You can either read it from here https://github.com/chainbound/bolt/blob/16152bde6129e91ab01a991bd87651cd37c2c2a5/bolt-contracts/src/contracts/BoltParametersV1.sol#L56 or hardcode it. As you prefer.

bolt-cli/src/contracts/mod.rs
thedevbirb: Makes sense! Ty

merklefruit (approved):
lgtm!

thedevbirb (approved):

PR #552, opened 2024-12-12
feat(sidecar): firewall delegation

### LINKED ISSUE #518: Firewall delegation support
Tracking issue for firewall delegation support in the sidecar. 

> Firewall delegation allows proposers to configure multiple “firewalls” to outsource filtering and pricing inclusion requests. It gets rid of the requirement for proposers to expose any public endpoints, but rather subscribe to any of the delegated firewalls for a stream of filtered and priced inclusion requests. Proposers running the sidecar are still responsible for signing commitments, submitting & verifying constraints, and building fallback blocks. They remain the single source of truth for valid commitments, which means we don’t have to deal with issues like split-brain and subjectivity. At the same time, the heavy lifting is outsourced to third parties with minimal added trust assumptions.

```[tasklist]
# Tasks
- [x] CLI flag `--firewall-rpc` for 1 or more values
- [x] Make commitments RPC optional and disable by default if `--firewall-rpc` is set
- [x] Initiate (durable) WebSocket connections with firewalls on `/api/v1/firewall_stream`
- [x] Multiplex incoming messages over single stream which yields origin as `oneshot::Sender` along with messages (for replies)
- [x] Wrap current commitment validation logic with with req/resp matching logic on uuids
```

thedevbirb:
Closes #518 

This PR introduces firewall delegation functionality into the sidecar. It does so by introducing a `firewall` module in the codebase which contains two components:
- a `CommitmentsReceiver`, which is an abstraction to connect to multiple firewall-ed websocket RPC servers and
forwards commitment request.
- a `CommitmentRequestProcessor` which, as the name suggests, is a processor of JSON-RPC requests of the Commitments-API. 

Some implementation notes:
- the websocket connection is authenticated via JWT. The JWT is made such that the claim contains all the data to identify a proposer opted into bolt, and encoded with the default `HS256` algorithm and empty encoding key.
- all messages sent in the websocket connection follow the JSON-RPC schema. This is allows great flexibility and allows the delegated operator to call other methods other than `bolt_requestInclusion`, such as `bolt_getVersion`.


bolt-sidecar/src/config/commitments.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// This port should be open on your firewall in order to receive external requests!
+    #[clap(long, env = "BOLT_SIDECAR_PORT", default_value = stringify!(DEFAULT_RPC_PORT))]
+    pub port: Option<u16>,
+    /// Comma-separated list of allowed RPC addresses to subscribe via websocket to receive
+    /// incoming commitments requests.
+    #[clap(
+        long,
+        env = "BOLT_SIDECAR_FIREWALL_RPC",
+        value_delimiter = ',',
+        conflicts_with("commitments_port")
```
merklefruit:
```suggestion
        conflicts_with("port")
```

bolt-sidecar/src/config/commitments.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Command-line options for signing constraint messages
+#[derive(Args, Deserialize, Debug, Clone)]
+#[clap(
+    group = ArgGroup::new("commitments-opts").required(true)
+        .args(&["commitments_port", "firewall_rpc"])
+)]
+pub struct CommitmentOpts {
+    /// Port to listen on for incoming JSON-RPC requests of the Commitments API.
+    /// This port should be open on your firewall in order to receive external requests!
+    #[clap(long, env = "BOLT_SIDECAR_PORT", default_value = stringify!(DEFAULT_RPC_PORT))]
```
merklefruit: with the current logic, won't this default value always default to using the commitments server even if you don't specify it?

bolt-sidecar/src/config/commitments.rs
```diff
@@ -0,0 +1,64 @@
...
+/// 8017 -> BOLT :)
+pub const DEFAULT_RPC_PORT: u16 = 8017;
+
+/// Command-line options for signing constraint messages
+#[derive(Args, Deserialize, Debug, Clone)]
+#[clap(
+    group = ArgGroup::new("commitments-opts").required(true)
+        .args(&["commitments_port", "firewall_rpc"])
+)]
+pub struct CommitmentOpts {
```
merklefruit, line 14: What do you think of this pattern to match on which method to use?

```env
BOLT_SIDECAR_COMMITMENT_SERVER_PORT=8017
BOLT_SIDECAR_USE_GATEWAYS=true
BOLT_SIDECAR_GATEWAY_URLS="1, 2, 3, ..."
```

```rust
if use_gateways {
	// start the firewall thing
} else {
	// start the server at port
}
```

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,548 @@
...
+    chain: Chain,
+    /// The URLs of the websocket servers to connect to.
+    urls: Vec<Url>,
+    /// The shutdown signal.
+    signal: Option<Pin<Box<dyn Future<Output = ()> + Send>>>,
+}
+
+impl Debug for CommitmentsFirewallRecv {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CommitmentsFirewallStream")
```
merklefruit:
```suggestion
        f.debug_struct("CommitmentsFirewallRecv")
```

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,548 @@
...
+                // Run the commitment processor indefinitely, reconnecting on failure.
+                commitment_request_processor.await
+            }
+            Err(e) => {
+                error!(?e, ?url, "failed to connect");
+            }
+        }
+
+        // Reconnect on failure
+        // TODO: add backoff
```
merklefruit: we could use the `retry_with_backoff` util here :)

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,548 @@
...
+            Box<
+                dyn Future<
+                        Output = Result<
+                            Result<SignedCommitment, CommitmentError>,
+                            oneshot::error::RecvError,
+                        >,
+                    > + Send,
+            >,
+        >,
+    >,
```
merklefruit: A good start could be to extract the type alias like
```rust
type PendingCommitmentResult = impl Future<Output = Result<SignedCommitment, CommitmentError>>;

pending_commitment_responses: FuturesUnordered<PendingCommitmentResult>
```

or similar (if you need pin<box as well)

Also I'm pretty sure we can get rid of the oneshot channel error and handle it in the poll implementation, as it can only happen when the sender is dropped aka if everything is shutting down

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,548 @@
...
+        let rpc_url = this.url.clone();
+
+        loop {
+            let mut progress = false;
+
+            // 1. Handle incoming websocket messages from the read stream.
+            while let Poll::Ready(maybe_message) = this.read_stream.poll_next_unpin(cx) {
+                progress = true;
+
+                match maybe_message {
```
merklefruit: overall the matching going on here feels a bit heavy!
can you make the control flow more linear by using something like:

```rust
let Some(Ok(message)) = maybe_message else {
	warn!("websocket connection with {rpc_url} dropped");
	return Poll::Ready(());
};

match message {
	Message::Text(...
	Message::Close(_) ...
```

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,548 @@
...
+                            // TODO: check whether this format is correct + handle errors.
+                            let message =
+                                Message::text(serde_json::to_string(&commitment).unwrap());
+                            // Add the message to the outgoing messages queue
+                            this.outgoing_messages.push_back(message);
+                        }
+                    }
+                    Err(e) => {
+                        error!(?e, "failed to receive commitment response");
+                    }
```
merklefruit: if this is a oneshot::error then we don't need to match on this, as it can only happen if the sender is dropped

merklefruit (changes requested):
left some initial comments; the actual request processor looks great so far but I will try it out in more detail later, also really nice seeing those tests :)

bolt-sidecar/src/config/commitments.rs
thedevbirb: Yes, this means if you spin up the sidecar without specifying any "gateway" or whatever it will be started in "server" mode.

bolt-sidecar/src/config/commitments.rs
thedevbirb: Why adding a "USE" flag when passing the `GATEWAY_URLS` env already conveys that you want to use delegation?
Also, this doesn't simplify the parsing logic of envs:
- what if "USE" is `true` but the list of gateway is empty?
- what if "USE" is unset but both options are provided?

bolt-sidecar/src/config/commitments.rs
merklefruit: No, it means if you spin up the sidecar specifying a list of gateways it will be started in server mode anyway because there is always a default value. that's what I understood from https://github.com/chainbound/bolt/pull/552/files#diff-d9dd450d19da416698fd8f0b096f157f4a2c36f99146757f589dde830b36adceR231

bolt-sidecar/src/api/commitments/firewall_recv.rs
thedevbirb: Right, but I'll keep the low and simplify the flow since these "dropped senders" errors are hard to catch.

bolt-sidecar/src/config/commitments.rs
thedevbirb: Right, good catch. Then I'll drop any default behaviour so that one of the two is mandatory and that's it.

merklefruit: this is why I proposed the USE_GATEWAYS variable, it's similar to how in geth you have to use 
`--rpc --rpc.addr 0.0.0.0 --rpc-port 8545` as in:

* `--rpc` enables the feature
* `--rpc.addr` and `.port` configure the rpc feature

in our case:
USE_GATEWAYS=true will dispatch the websocket provider etc
USE_GATEWAYS=false will always use the configured server port: default 8017 and can be configured

* if USE=true and the list is empty, we throw a runtime error because it's an invalid configuration
* if USE is unset but both options are provided we use the server port

wdyt?

bolt-sidecar/src/config/commitments.rs
merklefruit: responded in the comment above

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,622 @@
...
+/// The interval at which to send ping messages from connected clients.
+#[cfg(test)]
+const PING_INTERVAL: Duration = Duration::from_secs(3);
+#[cfg(not(test))]
+const PING_INTERVAL: Duration = Duration::from_secs(30);
+
+/// The maximum number of retries to attempt when reconnecting to a websocket server.
+const MAX_RETRIES: usize = 10;
+
+type ShutdownSignal = Pin<Box<dyn Future<Output = ()> + Send>>;
```
mempirate: Can also use `CancellationToken`s here: https://tokio.rs/tokio/topics/shutdown. Would actually be a good pattern to implement on all async tasks for graceful shutdown.

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,622 @@
...
+        // retry_with_backoff(MAX_RETRIES, fut)
+
+        // Reconnect on failure
+        // TODO: add backoff
+        warn!(?url, "connection lost. reconnecting...");
+        tokio::time::sleep(Duration::from_secs(5)).await;
+    }
+}
+
+type PendingCommitmentResult = dyn Future<Output = Result<Result<SignedCommitment, CommitmentError>, oneshot::error::RecvError>>
```
mempirate: See below, type can be simplified

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,622 @@
...
+                    let commitment_request_processor = CommitmentRequestProcessor::new(
+                        url.clone(),
+                        api_events_tx.clone(),
+                        write_sink,
+                        read_stream,
+                        ping_rx.resubscribe(),
+                        shutdown_rx.resubscribe(),
+                    );
+                    // Run the commitment processor indefinitely, reconnecting on failure.
+                    commitment_request_processor.await
```
mempirate: Was confused as to why this would trigger a reconnect, but the future itself returns a bool indicating that. Nice

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,622 @@
...
+                        // TODO: parse the text into a commitment request
+                        let request = CommitmentRequest::Inclusion(InclusionRequest::default());
+                        let event = CommitmentEvent { request, response: tx };
+
+                        if let Err(err) = this.api_events_tx.try_send(event) {
+                            error!(?err, "failed to forward commitment event to channel");
+                        }
+
+                        // add the pending response to this buffer for later processing
+                        this.pending_commitment_responses.push(rx.boxed());
```
mempirate: You can actually just push this `rx` as is without boxing it, and replace the type above with:

```rs
pending_commitment_responses:
        FuturesUnordered<oneshot::Receiver<Result<SignedCommitment, CommitmentError>>>,
```

This will also avoid a heap allocation

bolt-sidecar/src/api/commitments/firewall_recv.rs
```diff
@@ -0,0 +1,622 @@
...
+
+    fn poll(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        let this = self.get_mut();
+        let rpc_url = this.url.clone();
+
+        loop {
+            let mut progress = false;
```
mempirate: This pattern polls everything concurrently, with the same priority (which is fine). What you could also do is first handle messages with high priority, such as incoming requests / outbound commitments, where you `continue` until all of those messages are processed. You can then poll lower priority stuff like pings and shutdowns after. Here probably not needed but wanted to bring it up.

bolt-sidecar/src/config/commitments.rs
thedevbirb: I get the idea and I think it's works great if you're enabling features that are independent between each other.
In this context the "USE" flag will also have the additional semantics of invalidating another flag, which is fine but I'll try to avoid as possible.

Since when starting up the sidecar we have to distinguish between the two running modes with an `if` condition, what do you think to add a default there for the port? Like so:
```rs
        let api_events_rx = if let Some(urls) = opts.commitment_opts.gateway_urls.clone() {
            CommitmentsFirewallRecv::new(
                opts.commitment_opts.operator_private_key.clone(),
                opts.chain.chain,
                urls,
            )
            .run()
            .await
        } else {
            let port = opts.commitment_opts.port.unwrap_or(DEFAULT_RPC_PORT);
            // start the commitments api server
            let api_addr = format!("0.0.0.0:{}", port);
            let (api_events_tx, api_events_rx) = mpsc::channel(1024);
            CommitmentsApiServer::new(api_addr).run(api_events_tx, opts.limits).await;
            api_events_rx
        };
```

This would result in:
- you can either choose between `PORT` and `GATEWAY_URLS`
- if none of them are set, the default is `PORT=8017`.

bolt-sidecar/src/config/commitments.rs
mempirate: I would use `FIREWALL_RPCS` only, if that's set, disable HTTP server

bolt-sidecar/src/driver.rs
```diff
@@ -158,6 +159,7 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
         commitment_signer: ECDSA,
         fetcher: C,
     ) -> eyre::Result<Self> {
+        println!("opts {:#?}", opts);
```
mempirate: TODO

bolt-sidecar/src/driver.rs
```diff
@@ -227,10 +229,21 @@ impl<C: StateFetcher, ECDSA: SignerECDSA> SidecarDriver<C, ECDSA> {
...
         });
 
-        // start the commitments api server
-        let api_addr = format!("0.0.0.0:{}", opts.port);
-        let (api_events_tx, api_events_rx) = mpsc::channel(1024);
-        CommitmentsApiServer::new(api_addr).run(api_events_tx, opts.limits).await;
+        let api_events_rx = if let Some(port) = opts.commitment_opts.port {
+            // start the commitments api server
+            let api_addr = format!("0.0.0.0:{}", port);
+            let (api_events_tx, api_events_rx) = mpsc::channel(1024);
```
mempirate: Can we put `API_EVENTS_BUFFER_SIZE` in a const?

bolt-sidecar/src/primitives/commitment.rs
```diff
@@ -41,6 +41,13 @@ pub struct InclusionCommitment {
     signature: Signature,
 }
 
+impl InclusionCommitment {
+    /// Creates a new unchecked inclusion commitment
```
mempirate: Would add `#[cfg(test)]` because it should only be used in tests

mempirate (changes requested):
Great work already

bolt-sidecar/src/api/commitments/firewall_recv.rs
thedevbirb: Great catch, thanks!

bolt-sidecar/src/config/commitments.rs
merklefruit: I like this a lot more, thanks!

bolt-sidecar/src/api/commitments/firewall_recv.rs
merklefruit: This is prob overkill, but remember seeing it in the wild and it was cool :) 
https://github.com/superfly/corrosion/blob/edf04b0bc9bc6e2ae0dd42d69cb4368a2976eb24/crates/tripwire/src/tripwire.rs#L21

bolt-sidecar/src/config/commitments.rs
thedevbirb: Great, happy we've find a good design after some iterations :)

bolt-sidecar/src/api/commitments/firewall_recv.rs
thedevbirb: Reporting some of the internal discussion we had on the topic:
- there is already a priority expressed by continuously polling the `Stream` with `while let Poll::Ready(T)`.
- it's true that using `continue` allows for more granular logic and thanks for making me aware of it. I'll consider it in case of further optimizations of the polling logic in the future.

bolt-sidecar/src/api/commitments/delegation/mod.rs
```diff
@@ -0,0 +1,6 @@
+/// The JWT used for authenticating a websocket connection.
+pub mod jwt;
+/// The processor of incoming commitments request via a websocket connection.
+pub mod processor;
+/// The Commitments receiver.
+pub mod receiver;
```
merklefruit: nit

```suggestion
/// The JWT used for authenticating a websocket connection.
pub mod jwt;

/// The processor of incoming commitments request via a websocket connection.
pub mod processor;

/// The Commitments receiver.
pub mod receiver;
```

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// The websocket writer sink.
+    write_sink: SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>,
+    /// The websocket reader stream.
+    read_stream: SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>,
+    /// The receiver for keep-alive ping messages.
+    ping_rx: broadcast::Receiver<()>,
+    /// The receiver for shutdown signals.
+    shutdown_rx: broadcast::Receiver<()>,
+    /// The collection of pending commitment requests responses, sent with [api_events_tx].
+    /// NOTE: Is there a better way to avoid this monster type?
```
merklefruit: the type doesn't look too bad to me!

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -0,0 +1,232 @@
...
+            ping_rx,
+            shutdown_rx,
+            pending_commitment_responses: FuturesUnordered::new(),
+            outgoing_messages: VecDeque::new(),
+        }
+    }
+}
+
+impl Future for CommitmentRequestProcessor {
+    // The output of this future is a boolean indicating whether reconnection is needed.
```
merklefruit: not a boolean anymore I guess, but the enum is self explanatory :)

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -0,0 +1,232 @@
...
+                // Check if the write sink is able to receive data.
+                match this.write_sink.poll_ready_unpin(cx) {
+                    Poll::Ready(Ok(())) => {
+                        progress = true;
+
+                        // Try to send the message to the sink, for later flushing.
+                        if let Err(e) = this.write_sink.start_send_unpin(message) {
+                            error!(?e, ?rpc_url, "failed to send message to websocket write sink");
+                            // NOTE: Should we return here?
+                            // return Poll::Ready(());
```
merklefruit: I would just continue, the log is good enough

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -0,0 +1,232 @@
...
+                    }
+                    Poll::Pending => {
+                        // Put the message back and try again later
+                        this.outgoing_messages.push_front(message);
+                        break;
+                    }
+                    Poll::Ready(Err(e)) => {
+                        error!(?e, "sink error while sending message to websocket");
+                        // NOTE: Should we return here?
+                        // return Poll::Ready(());
```
merklefruit: same

bolt-sidecar/src/api/commitments/delegation/receiver.rs
```diff
@@ -0,0 +1,511 @@
...
+use reqwest::Url;
+
+use crate::api::commitments::delegation::processor::CommitmentRequestProcessor;
+use crate::api::commitments::server::CommitmentEvent;
+use crate::common::backoff::retry_with_backoff_if;
+use crate::common::secrets::EcdsaSecretKeyWrapper;
+use crate::config::chain::Chain;
+
+use super::jwt::ProposerAuthClaims;
+use super::processor::InterruptReason;
```
merklefruit: nit: why are imports such as `crate::` not under the same block? my formatter always groups them...
smaller nit: `std` imports should be grouped and sit at the top separated by a new line also

bolt-sidecar/src/api/commitments/delegation/receiver.rs
```diff
@@ -0,0 +1,511 @@
...
+            loop {
+                ping_interval.tick().await;
+                if let Err(err) = ping_tx.send(()) {
+                    error!(?err, "internal error while sending ping task")
+                }
+            }
+        });
+
+        let signal = self.signal.take();
+        if let Some(signal) = signal {
```
merklefruit: nit
```suggestion
        if let Some(signal) = self.signal.take() {
```

bolt-sidecar/src/api/commitments/mod.rs
```diff
@@ -1,12 +1,5 @@
...
-mod headers;
-/// JSON-RPC helper types and functions.
-mod jsonrpc;
-/// The commitments-API middleware.
-mod middleware;
-/// The commitments-API JSON-RPC server implementation.
+/// The logic to accept commitments via firewall delegation, where a proposer specifies a list of
+/// RPC URLs from which it will accept commitments via a websocket connection.
+pub mod delegation;
+/// The Commitments-API server implementation.
```
merklefruit:
```suggestion
pub mod delegation;

/// The Commitments-API server implementation.
```

bolt-sidecar/src/api/commitments/server/mod.rs
```diff
@@ -1,3 +1,16 @@
+/// The commitments-API request handlers.
+mod handlers;
+/// The commitments-API headers and constants.
+mod headers;
+/// JSON-RPC helper types and functions.
+mod jsonrpc;
+/// The commitments-API middleware.
+mod middleware;
+/// The commitments-API specification and errors.
+pub mod spec;
```
merklefruit:
```suggestion
/// The commitments-API request handlers.
mod handlers;

/// The commitments-API headers and constants.
mod headers;

/// JSON-RPC helper types and functions.
mod jsonrpc;

/// The commitments-API middleware.
mod middleware;

/// The commitments-API specification and errors.
pub mod spec;
```

bolt-sidecar/src/common/backoff.rs
```diff
@@ -19,6 +19,24 @@ where
     Retry::spawn(backoff, fut).await
 }
 
+/// Retry a future with exponential backoff and jitter if the error matches a condition.
+pub async fn retry_with_backoff_if<F, T, E>(
```
merklefruit, line 41: wait, didn't we have this already somewhere? or was it on the rpc?

bolt-sidecar/src/config/chain.rs
```diff
@@ -97,6 +97,16 @@ pub enum Chain {
 }
 
 impl Chain {
+    /// Get the chain ID for the given chain.
+    pub const fn id(&self) -> u64 {
```
merklefruit, line 114: we already have `ChainConfig::chain_id()`. if you want to keep this, please update that function to just call `chain.id()` instead of matching again. I don't mind either one option

bolt-sidecar/src/config/mod.rs
```diff
@@ -119,7 +113,16 @@ impl Opts {
...
-        Ok(Self::parse())
+        let opts = Self::parse();
+        opts.validate()?;
+        Ok(opts)
+    }
+
+    fn validate(&self) -> eyre::Result<()> {
+        if self.commitment_opts.firewall_rpcs.is_some() && self.limits.min_priority_fee.is_some() {
+            bail!("min priority fee cannot be set when using the firewall RPCs");
+        }
```
merklefruit: Why is this the case? just curious

bolt-sidecar/src/driver.rs
```diff
@@ -122,7 +124,7 @@ impl SidecarDriver<StateClient, PrivateKeySigner> {
         let keystore_signer = SignerBLS::Keystore(keystore);
 
         // Commitment responses are signed with a regular Ethereum wallet private key.
-        let commitment_key = opts.commitment_private_key.0.clone();
+        let commitment_key = opts.commitment_opts.operator_private_key.0.clone();
```
merklefruit, line 128: fwiw, i like the new naming, commitment_opts.operator_private_key is very clear, nice improvement!

bolt-sidecar/src/primitives/signature.rs
```diff
@@ -0,0 +1,101 @@
...
+use serde::{de, Deserialize, Deserializer, Serialize};
+
+/// Error type for signature errors.
+#[derive(Debug, thiserror::Error)]
+#[error("Invalid signature")]
+pub struct SignatureError;
+
+/// A wrapper over [alloy::primitives::PrimitiveSignature] that implements serde serialization and
+/// deserialization.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, Deref, DerefMut, From, FromStr)]
```
merklefruit, line 17: nice derive_more usage

merklefruit (changes requested):
Second round of review, left mostly nits and some questions, great job!

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -26,6 +29,34 @@ pub enum InterruptReason {
...
+
+impl Future for PendingCommitmentResponse {
+    type Output = (Uuid, Result<Result<SignedCommitment, CommitmentError>, RecvError>);
+    fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        let id = this.id;
+
+        match this.response.poll_unpin(cx) {
+            Poll::Ready(Ok(response)) => Poll::Ready((id, Ok(response))),
+            Poll::Ready(Err(e)) => Poll::Ready((id, Err(e))),
```
merklefruit: can be coerced to `Poll::Ready(res) => Poll::Ready((id, res))`

bolt-sidecar/src/api/commitments/delegation/processor.rs
```diff
@@ -98,13 +128,20 @@ impl Future for CommitmentRequestProcessor {
...
                 let Ok(result_commitment) = response else {
                     error!("failed to receive commitment response. dropped sender");
                     continue;
                 };
 
                 if let Ok(commitment) = result_commitment {
                     trace!(?rpc_url, ?commitment, "received commitment response");
+                    let commitment = commitment
+                        .into_inclusion_commitment()
+                        .expect("no other commitment types")
```
merklefruit: What do you think of using a match statement instead of this `into` function?
that way we will be prompted to update the code when we add a new enum variant

bolt-sidecar/src/common/backoff.rs
thedevbirb: This is the `if` version, but it might be we have it already and I have to pull from upstream

bolt-sidecar/src/config/mod.rs
thedevbirb: Because pricing logic is delegated and the sidecar isn't enforcing minimums

bolt-sidecar/src/api/commitments/delegation/receiver.rs
thedevbirb: I'm not using nightly formatter on my LSP. A run of `just fmt bolt-sidecar` solved it!

bolt-sidecar/src/config/mod.rs
mempirate: Will be changed slightly in #599, as we added a min profit parameter

bolt-sidecar/src/api/firewall/receiver.rs
```diff
@@ -0,0 +1,519 @@
...
+};
+
+/// The interval at which to send ping messages from connected clients.
+#[cfg(test)]
+const PING_INTERVAL: Duration = Duration::from_secs(3);
+#[cfg(not(test))]
+const PING_INTERVAL: Duration = Duration::from_secs(30);
+
+/// The maximum number of retries to attempt when reconnecting to a websocket server.
+const MAX_RETRIES: usize = 1000;
```
estensen: shouldn't we retry forever?

bolt-sidecar/src/api/firewall/receiver.rs
```diff
@@ -0,0 +1,519 @@
...
+                            None,
+                            signer.clone(),
+                        )
+                        .to_jwt()
+                        .expect("failed to produce JWT");
+
+                        let api_events_tx = api_events_tx.clone();
+                        let ping_rx = ping_rx.resubscribe();
+                        let shutdown_rx = shutdown_rx.resubscribe();
+
```
estensen: nit: extract to function to make the function shorter?

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -0,0 +1,564 @@
...
+            let commitment_request_processor = CommitmentRequestProcessor::new(
+                url,
+                state,
+                api_events_tx.clone(),
+                write_sink,
+                read_stream,
+                ping_rx.resubscribe(),
+                shutdown_rx.resubscribe(),
+            );
+            // Run the commitment processor indefinitely, reconnecting on failure.
```
estensen: I don't understand how this reconnection works. Could you please elaborate?

estensen: could be handled in a future PR, this one is already huge, but I would make sure reconnection with exponential backoff is present

bolt-sidecar/src/api/commitments/firewall/processor.rs
```diff
@@ -0,0 +1,350 @@
...
+                }
+                Err(TryRecvError::Closed) => {
+                    error!("shutdown channel closed");
+                }
+                _ => {}
+            }
+
+            // Incoming work tasks
+
+            // 5. Handle incoming websocket messages from the read stream.
```
estensen, line 206: consider adding rate-limiting to protect against rogue firewalls

estensen: and what about adding a circuit-breaker in case things to haywire?

bolt-sidecar/src/api/commitments/firewall/processor.rs
```diff
@@ -0,0 +1,350 @@
...
+                            Err(e) => Err(e),
+                            Ok(request) => {
+                                let id = request.id;
+
+                                match request.method.as_str() {
+                                    GET_VERSION_METHOD => Ok(JsonResponse {
+                                        id: Some(Value::String(id.to_string())),
+                                        result: Value::String(BOLT_SIDECAR_VERSION.clone()),
+                                        ..Default::default()
+                                    }),
```
estensen: from [spec](https://www.jsonrpc.org/specification)

> jsonrpc
    A String specifying the version of the JSON-RPC protocol. MUST be exactly "2.0". 

```suggestion
                                    GET_VERSION_METHOD => Ok(JsonResponse {
                                        jsonrpc: "2.0".to_string(),
                                        id: Some(Value::String(id.to_string())),
                                        result: Value::String(BOLT_SIDECAR_VERSION.clone()),
                                        ..Default::default()
                                    }),
```

bolt-sidecar/src/api/commitments/firewall/receiver.rs
thedevbirb: Sorry, the comment is stale/misleading and I will remove it, but you can see that retries exponential backoff are already implemented in `CommitmentsReceiver::run`.

bolt-sidecar/src/api/commitments/firewall/processor.rs
thedevbirb: The `..Default::default()` already takes care of that

bolt-sidecar/src/api/commitments/firewall/processor.rs
thedevbirb: The idea of firewall delegation is to not worry about rate-limiting and filtering, which is done by the trusted operator to which you're connected to. 
Can you provide an example about the second comment?

estensen: I understand that firewalls at least needs to be semi-trusted, but isn't it bad for the network if an operator is hacked and they DoS the network? With rate-limiters we can make sure a single firewall can't do too much damage. And with a circurit-breaker we can turn off accepting preconfs if there's a catastrofic event.

But I understand if we don't wanna focus on this now. Can create an issue and then we can consider it down the line.

thedevbirb: I get the reasoning, let's think about it at a later stage! cc @mempirate

bolt-sidecar/src/api/firewall/receiver.rs
thedevbirb: I have not made a decision on this yet. I wonder if at some point it would be better to panic.

bolt-sidecar/src/api/commitments/firewall/processor.rs
mempirate: Yeah I agree with Havard, we need to have a defense in depth mindset about these things. Trusting an external entity is fine until it's not, issues can arise not just from malicious intent but also hacks, bugs, and other stuff. Always remember that this is running on validator hardware.

We can create a follow-up issue though, doesn't need to be implemented right away.

mempirate: https://github.com/chainbound/bolt/issues/646

bolt-sidecar/src/api/commitments/firewall/jwt.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///
+    /// Example digest: `keccak256("http://localhost:8545" + 1 + 1734104048)`
+    ///
+    ///
+    /// The signature must be performed with the operator private key.
+    signature: String,
+}
+
+impl ProposerAuthClaims {
+    /// Creates a new instance of the [ProposerAuthClaim] and creates a signature with
```
merklefruit:
```suggestion
    /// Creates a new instance of the [ProposerAuthClaims] and creates a signature with
```

bolt-sidecar/src/api/commitments/firewall/processor.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// The websocket writer sink.
+    write_sink: SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>,
+    /// The websocket reader stream.
+    read_stream: SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>,
+    /// The receiver for keep-alive ping messages.
+    ping_rx: broadcast::Receiver<()>,
+    /// The receiver for shutdown signals.
+    shutdown_rx: broadcast::Receiver<()>,
+    /// The collection of pending commitment requests responses, sent with [api_events_tx].
+    /// NOTE: Is there a better way to avoid this monster type?
```
merklefruit: rm this line, the type looks great now

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -0,0 +1,616 @@
...
+const PING_INTERVAL: Duration = Duration::from_secs(30);
+
+/// The maximum number of retries to attempt when reconnecting to a websocket server.
+const MAX_RETRIES: usize = 1000;
+
+/// The maximum messages size to receive via websocket connection, in bits, set to 32MiB.
+///
+/// It is enough to account for a commitment request with 6 blobs and the largest
+/// memory-consuming transactions you can create. Reference: https://xn--2-umb.com/22/eth-max-mem/
+const MAX_MESSAGE_SIZE: usize = 16 << 23;
```
merklefruit: isn't 16 x 2^23 equal to 128MiB?

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -0,0 +1,616 @@
...
+    /// events.
+    pub fn run(self) -> mpsc::Receiver<CommitmentEvent> {
+        // mspc channel where every websocket connection will send commitment events over its own
+        // tx to a single receiver.
+        let (api_events_tx, api_events_rx) = mpsc::channel(self.urls.len() * 2);
+        let (ping_tx, ping_rx) = broadcast::channel::<()>(1);
+        let (shutdown_tx, shutdown_rx) = broadcast::channel::<()>(1);
+
+        PingTicker::new(PING_INTERVAL).spawn(ping_tx);
+        ShutdownTicker::new(self.signal).spawn(shutdown_tx);
```
merklefruit, line 121: I really like this, well done!

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -0,0 +1,616 @@
...
+            // able to handle incoming message requests.
+            let commitment_request_processor = CommitmentRequestProcessor::new(
+                url,
+                state,
+                api_events_tx.clone(),
+                write_sink,
+                read_stream,
+                ping_rx.resubscribe(),
+                shutdown_rx.resubscribe(),
+            );
```
merklefruit: nit: why not `.split()` the stream in `CommitmentRequestProcessor::new` instead of here?

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -0,0 +1,616 @@
...
+            // For each opened connection, create a new commitment processor
+            // able to handle incoming message requests.
+            let commitment_request_processor = CommitmentRequestProcessor::new(
+                url,
+                state,
+                api_events_tx.clone(),
+                write_sink,
+                read_stream,
+                ping_rx.resubscribe(),
+                shutdown_rx.resubscribe(),
```
merklefruit: why do these need to be cloned? I don't see a loop here, what am I missing?

bolt-sidecar/src/primitives/jsonrpc.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct JsonRpcRequestUuid {
+    /// The JSON-RPC version string. MUST be "2.0".
+    pub jsonrpc: String,
+    /// The method string.
+    pub method: String,
+    /// Optional ID.
+    pub id: Uuid,
+    /// The parameters object.
+    pub params: Vec<Value>,
+}
```
merklefruit, line 29: why is this needed? can't we reuse Identified<JsonRpcRequest, Uuid> ?

bolt-sidecar/src/primitives/misc.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl<T, S> DerefMut for Signed<T, S> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
```
merklefruit: please use derive_more

bolt-sidecar/src/primitives/misc.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl<T, I: Copy> DerefMut for Identified<T, I> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
```
merklefruit: same as above

merklefruit (changes requested):
Great work once again! I went through everything and just left a few small nits, overall the quality is very high and the refactors you added are super nice too :)

bolt-sidecar/src/api/commitments/firewall/processor.rs
thedevbirb: ops, sorry again

bolt-sidecar/src/api/commitments/firewall/receiver.rs
thedevbirb: You're right, `16 << 23` is 128 MiB. I don't remember why I shifted it by 23, writing `32 << 20` is much simpler. Going to change it

bolt-sidecar/src/api/commitments/firewall/receiver.rs
thedevbirb: Good idea!

bolt-sidecar/src/api/commitments/firewall/receiver.rs
thedevbirb: Good catch, it is indeed not needed

bolt-sidecar/src/primitives/jsonrpc.rs
thedevbirb: Because `Identified` creates a struct with `id` and `inner` field, which doesn't exactly fit a JSON-RPC object. I could have implemented a generic version of `JsonRpcRequest` over `id` but it seemed a bit overkill, also because I want it required in this case.

This made me realize that a procedural macro that adds an "id" field might be better than the `Identified` type, let's see...

merklefruit: but since `inner` has a `#[flatten]` attribute it shouldn't matter for deserialization, right?
no need for proc macros IMO :D

thedevbirb: Hmm true but `JsonRpcRequest` as already an `id` field, as such if you create a struct `Identified<JsonRpcRequest, Uuid>` when serializing you will have two `id` fields. This is how it would be serialized:
```
"{\"jsonrpc\":\"\",\"method\":\"\",\"id\":null,\"params\":[],\"id\":\"019445a4-988d-7eb1-b6c9-444ef07213e8\"}"
```

scripts/kurtosis_config.yaml
```diff
@@ -26,8 +26,10 @@ mev_params:
   # Bolt-specific images:
   # Adding the `bolt_boost_image` will start the devnet with Bolt-Boost
   # instead of MEV-Boost by Flashbots
-  bolt_boost_image: ghcr.io/chainbound/bolt-boost:0.1.0
+  # bolt_boost_image: ghcr.io/chainbound/bolt-boost:0.1.0
   bolt_sidecar_image: ghcr.io/chainbound/bolt-sidecar:0.1.0
+  # bolt_sidecar_extra_params:
+  #   firewall_rpcs: "ws://host.docker.internal:8015/api/v1/firewall_stream"
```
estensen, line 32: Do you still want this to be commented out?

estensen (approved):
Great work!

bolt-sidecar/src/primitives/jsonrpc.rs
merklefruit: Ah you're right, I didn't notice

scripts/kurtosis_config.yaml
thedevbirb: Yes, the firewall-ed setup should not be the default because it requires a bolt-rpc ready to run, which isn't publicly available as of now.
I think it's still useful to leave this commented so that's easy to turn it on if needed, like it has been done between MEV-Boost fork and bolt-boost

mempirate (approved):
LGTM 🚀

mempirate:
@thedevbirb to do in follow-up after testing on Holesky: update `testnets/holesky` folder with configs and guides

merklefruit (approved):
🚀

PR #554, opened 2024-12-12
chore(bolt-boost): bump alloy

estensen:


bolt-boost/Cargo.toml
```diff
@@ -43,8 +43,9 @@ alloy = { version = "0.5.4", features = [
 alloy-rlp = "0.3.9"
 
 # commit-boost
-cb-common = { git = "https://github.com/commit-boost/commit-boost-client", tag = "v0.4.0" }
-cb-pbs = { git = "https://github.com/commit-boost/commit-boost-client", tag = "v0.4.0" }
+# pinned to rev be able to bump alloy
+cb-common = { git = "https://github.com/commit-boost/commit-boost-client", rev = "0f8f69b" }
```
mempirate, line 47: Will dependabot autobump these if it sees more recent tags?

estensen: No, but it wasn't bumping alloy when version was set either 🤷🏻

merklefruit (approved):
lgtm

PR #571, opened 2024-12-16
chore(bolt-sidecar): PooledTransactionsElement -> PooledTransaction

estensen:
`PooledTransactionsElement` is deprecated

### DIFF
--- a/bolt-sidecar/src/primitives/transaction.rs
+++ b/bolt-sidecar/src/primitives/transaction.rs
@@ -28,10 +28,10 @@ impl TransactionExt for PooledTransaction {
     fn value(&self) -> U256 {
         match self {
             Self::Legacy(transaction) => transaction.tx().value,
-            Self::Eip2930(transaction) => transaction.tx().value,
             Self::Eip1559(transaction) => transaction.tx().value,
+            Self::Eip2930(transaction) => transaction.tx().value,
             Self::Eip4844(transaction) => transaction.tx().tx().value,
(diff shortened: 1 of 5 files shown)


bolt-sidecar/src/primitives/transaction.rs
```diff
@@ -123,15 +130,31 @@ impl std::ops::DerefMut for FullTransaction {
...
+            PooledTransaction::Legacy(tx) => tx.into(),
+            PooledTransaction::Eip2930(tx) => tx.into(),
+            PooledTransaction::Eip1559(tx) => tx.into(),
+            PooledTransaction::Eip4844(tx) => {
+                let sig = *tx.signature();
+                let hash = *tx.hash();
+                let inner_tx = tx.into_parts().0.into_parts().0;
+                Signed::new_unchecked(inner_tx, sig, hash).into()
+            }
+            _ => unimplemented!(),
```
merklefruit: can you remove this? so that if we get a new transaction type we know what we need to update

merklefruit (changes requested):
Nice! Hopefully this doesn't change anymore :D

bolt-sidecar/src/primitives/transaction.rs
estensen: good catch, fixed this two other places as well

thedevbirb (approved):

merklefruit (approved):

PR #551, opened 2024-12-12
feat(contracts): `BoltManagerV3`

merklefruit:
This PR introduces a new version of BoltManager that allows for fetching operator data through view methods.

### New functionality

* `function updateOperatorRPC(string calldata rpc) external` to update an operator's RPC endpoint
* `function getOperatorData(address operator) external view returns (EnumerableMapV3.Operator memory operatorData)` to get the current operator data, like middleware, start time and rpc endpoint.
* `function getAllOperatorsData() external view returns (EnumerableMapV3.Operator[] memory operatorData)` to get all operators data in a single call.

- closes #397
- closes #408

PR #577, opened 2024-12-16
feat(contracts): `BoltValidatorsV3`

### LINKED ISSUE #409: chore(contracts): add a way to read registered validators by range
`BoltValidators` has a `function getAllValidators() public view returns (ValidatorInfo[] memory)`
that fails to execute on regular RPCs because of simulation gas limits.

One way to work around this is to increase the gas limit (e.g. with `--rpc.gascap` on geth).

However we could introduce a function that reads only a slice of N validators from a start_index.
something like `getValidatorsPaginated`.

merklefruit:
NOTE: after #551 

New version with `getValidatorsByRange`

- closes #409


mempirate:
@merklefruit I think we can close this given the registry revamp. Wdyt?

merklefruit:
> @merklefruit I think we can close this given the registry revamp. Wdyt?

Yes, I agree

PR #578, opened 2024-12-16
fix(bolt-sidecar): retry when EL is unavailable

### LINKED ISSUE #369: qol(sidecar): driver initialization fails if execution client hangs
When starting bolt-sidecar with an EL node that's still syncing or is otherwise unreachable, we panic during driver initialization:

```
2024-11-08T15:32:13.710198Z DEBUG bolt_sidecar::chain_io::manager: fetching 1 proposer statuses for chunk 1 of 1
Error: Failed to initialize the sidecar driver with local signer: error sending request for url (http://localhost:8545/)

Caused by:
   0: error sending request for url (http://localhost:8545/)
   1: client error (Connect)
   2: tcp connect error: Connection refused (os error 111)
   3: Connection refused (os error 111)

Location:
    /app/src/chain_io/manager.rs:116:33
```

It would be cleaner to instead output a clear message explaining that the EL node is unreachable and retrying if possible.
We could reuse the `retry_with_backoff` util that was recently added perhaps.

Just to be clear, this _is an error_ but it should be handled more gracefully.

estensen:
Fix https://github.com/chainbound/bolt/issues/369

### DIFF
--- a/bolt-sidecar/src/chain_io/manager.rs
+++ b/bolt-sidecar/src/chain_io/manager.rs
@@ -2,24 +2,27 @@ use alloy::{
     primitives::Address,
     providers::{ProviderBuilder, RootProvider},
     sol,
-    transports::http::Http,
+    transports::{http::Http, RpcError},
 };
+use alloy_contract::Error;
 use ethereum_consensus::primitives::BlsPublicKey;
 use eyre::{bail, Context};
 use reqwest::{Client, Url};
 use serde::Serialize;
 
-use tracing::debug;
+use tracing::{debug, warn};
 use BoltManagerContract::{
     BoltManagerContractErrors, BoltManagerContractInstance, ProposerStatus, ValidatorDoesNotExist,
 };
 
-use crate::config::chain::Chain;
+use crate::{common::backoff::retry_with_backoff, config::chain::Chain};
 
 use super::utils::{self, CompressedHash};
 
 /// Maximum number of keys to fetch from the EL node in a single query.
 const MAX_CHUNK_SIZE: usize = 100;
+/// Maximum number of retries for EL node connection attempts
+const MAX_RETRIES: usize = 5;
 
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/chain_io/manager.rs
```diff
@@ -70,15 +73,38 @@ impl BoltManager {
...
+                                let decoded_error = utils::try_parse_contract_error(err)
+                                    .wrap_err("Failed to fetch proposer statuses from EL client")?;
+
+                                bail!(generate_bolt_manager_error(
+                                    decoded_error,
+                                    commitment_signer_pubkey,
+                                ))
+                            }
+                        }
+                    }
```
thedevbirb: Here you should be able to avoid a level of indentation by matching the inner enum directly, like so:
```rs
Err(Error::TransportError(RpcError::Transport(err)) => { /* ... */ }
Err(_) => { /* ... */ }
```

thedevbirb (approved):

bolt-sidecar/src/chain_io/manager.rs
```diff
@@ -70,15 +73,38 @@ impl BoltManager {
...
 
-            let returndata = match self.0.getProposerStatuses(hashes_chunk).call().await {
-                Ok(returndata) => returndata,
-                Err(error) => {
-                    let decoded_error = utils::try_parse_contract_error(error)
-                        .wrap_err("Failed to fetch proposer statuses from EL client")?;
-
-                    bail!(generate_bolt_manager_error(decoded_error, commitment_signer_pubkey));
+            // Retry the contract call with exponential backoff
+            let returndata = retry_with_backoff(MAX_RETRIES, || async {
```
estensen: Need `retry_with_backoff_if`, now it always retries

thedevbirb (approved):

PR #580, opened 2024-12-16
chore: parse ether value with units

### LINKED ISSUE #410: feat(cli): parse amounts with measurements
### Describe the feature

We think the best UX for parsing amounts in the `bolt` CLI would be using also the units of measure, like `1.2ether`, `0.5gwei` or `1000000wei`. For reference, this is how `cast send` does it.

merklefruit:
closes #410

### DIFF
--- a/bolt-cli/src/common/mod.rs
+++ b/bolt-cli/src/common/mod.rs
@@ -1,11 +1,14 @@
 use std::{fs, io::Write, path::PathBuf, str::FromStr};
 
 use alloy::{
-    contract::Error as ContractError, primitives::Bytes, sol_types::SolInterface,
+    contract::Error as ContractError,
+    dyn_abi::DynSolType,
+    primitives::{Bytes, U256},
+    sol_types::SolInterface,
     transports::TransportError,
 };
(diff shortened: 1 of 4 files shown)


bolt-cli/src/common/mod.rs
```diff
@@ -27,6 +30,23 @@ pub mod hash;
...
+pub fn parse_ether_value(value: &str) -> Result<U256> {
+    Ok(if value.starts_with("0x") {
+        U256::from_str_radix(value, 16)?
+    } else {
+        DynSolType::coerce_str(&DynSolType::Uint(256), value)?
+            .as_uint()
+            .wrap_err("Could not parse ether value from string")?
+            .0
+    })
+}
```
merklefruit: this is taken from `cast`

thedevbirb: Can you add a reference to the code snippet?

estensen (approved):

bolt-cli/src/common/mod.rs
merklefruit: done!

thedevbirb (approved):

PR #582, opened 2024-12-16
fix(cli): return non-ABI error responses

merklefruit:
Fixes the weird error response:

```text
Error: type check failed for "BoltValidatorsErrors" with data: 

Caused by:
    type check failed for "BoltValidatorsErrors" with data: 

Location:
    src/commands/validators.rs:83:39
```

when receiving a response from the RPC that's not necessarily an ABI-decodable error (for instance, "insufficient funds")

### DIFF
--- a/bolt-cli/src/common/mod.rs
+++ b/bolt-cli/src/common/mod.rs
@@ -7,7 +7,7 @@ use alloy::{
 use ethereum_consensus::crypto::PublicKey as BlsPublicKey;
 use eyre::{Context, Result};
 use serde::Serialize;
-use tracing::info;
+use tracing::{error, info};
 
 /// BoltManager contract bindings.
 pub mod bolt_manager;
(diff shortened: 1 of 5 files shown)


bolt-cli/src/common/mod.rs
```diff
@@ -69,12 +69,19 @@ pub fn write_to_file<T: Serialize>(out: &str, data: &T) -> Result<()> {
 pub fn try_parse_contract_error<T: SolInterface>(error: ContractError) -> Result<T, ContractError> {
     match error {
         ContractError::TransportError(TransportError::ErrorResp(resp)) => {
-            let data = resp.data.unwrap_or_default();
+            let data = resp.data.clone().unwrap_or_default();
```
thedevbirb: Do we need this `clone`?

thedevbirb (approved):

estensen (approved):

bolt-cli/src/common/mod.rs
merklefruit: Done in the last commit

mempirate (approved):

PR #583, opened 2024-12-16
chore: add just test task

estensen:


justfile
```diff
@@ -2,6 +2,16 @@
...
   @just --list --unsorted
 
+# run tests in all packages using nextest
+test:
+    @just test-package bolt-cli
+    @just test-package bolt-boost
+    @just test-package bolt-sidecar
+
+# run tests in a specific package using nextest
+test-package package:
```
estensen: think it's better that `just test` tests all packages than this, but open for other suggestions on what to call it

thedevbirb: If you want you can have a command for them all!
```justfile
# Run tests in all packages or a specific package
test package='': 
    @if [[ "{{package}}" == "" ]]; then \
        just test-package bolt-cli; \
        just test-package bolt-boost; \
        just test-package bolt-sidecar; \
    else \
        just test-package {{package}}; \
    fi
```
Make sure this works, I know it is possible but this is gpt-generated :D

estensen: ty, it worked

merklefruit (approved):

PR #484, opened 2024-11-28
feat(ci): add security audit, rename files

mempirate:
Closes #415

Issue #415, opened 2024-11-21
Add rustsec action


mempirate:
https://github.com/rustsec/audit-check

Also more clippy lints

PR #591, opened 2024-12-17
chore: add audit action

estensen:
Solve https://github.com/chainbound/bolt/issues/415

Alternative to https://github.com/chainbound/bolt/pull/484/files where https://github.com/actions-rs/audit-check is deprecated.

This Action doesn't have too many users, but used by rust-lang, cloudflare and tokio-rs, so consider it ok.
https://grep.app/search?q=rustsec/audit-check

Run on PRs where deps change, will add follow-up where it's also run periodically.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,14 @@
+name: Security audit
+on:
+  push:
+    paths: 
+      - '**/Cargo.toml'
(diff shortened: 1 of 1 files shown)


estensen:
Need to verify that `cargo audit` passes before merging

mempirate:
Closing in favor of #484

PR #592, opened 2024-12-17
fix(sidecar): don't send empty delegations and warn about it

### LINKED ISSUE #587: Warn/Error when sidecar is sending empty delegation list
Consider this logic in the sidecar, run when the beacon node is sending `register_validators` requests.

https://github.com/chainbound/bolt/blob/eed9cec9b644632550479f05823b4487d3ed1ed6/bolt-sidecar/src/client/constraints.rs#L111-L148

If `filtered_delegations` is empty, this is probably a misconfiguration on the user side and we should either panic or print an error/warn message.
Such scenario happens when a user is enabling delegations done with a set of validator public keys that doesn't completely match the set of validator keys used to run the beacon client.

It might be intended behaviour if for example the user wants to run the sidecar but wants to issue commitments only with a subset of its public keys, and just proxy request to regular PBS otherwise.

thedevbirb:
Closes #587

### DIFF
--- a/bolt-sidecar/src/client/constraints.rs
+++ b/bolt-sidecar/src/client/constraints.rs
@@ -8,7 +8,7 @@ use ethereum_consensus::{
     deneb::mainnet::SignedBlindedBeaconBlock, Fork,
 };
 use reqwest::Url;
-use tracing::error;
+use tracing::{error, span_enabled, trace, warn, Level};
 
 use crate::{
     api::{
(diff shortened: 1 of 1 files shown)


bolt-sidecar/src/client/constraints.rs
```diff
@@ -140,6 +140,24 @@ impl BuilderApi for ConstraintsClient {
             .cloned()
             .collect::<Vec<_>>();
 
+        if filtered_delegations.is_empty() {
+            warn!("No delegations found for the incoming validator registrations");
+            // Works also with directives like `RUST_LOG=bolt_sidecar=trace`
+            if span_enabled!(Level::TRACE) {
```
merklefruit, line 146: nice trick

merklefruit (approved):

bolt-sidecar/src/client/constraints.rs
mempirate: Very nice trick

PR #593, opened 2024-12-18
feat(bolt-sidecar): implement pricing models for preconfs

estensen:
Implement pricing model from https://research.lido.fi/t/a-pricing-model-for-inclusion-preconfirmations/9136

Some limitations that we need to consider before using this model, is that the price of a preconf is both dependent on how much gas has already been preconfirmed for a slot and how much gas a preconf tx will consume:

**Price race**
If preconfirmations P1 and P2 race the, latter will use an outdated price because the price depends on how much gas has been preconfirmed. We don't want P2 to be rejected, as this would result in poor UX. It will also limit how many tx we can preconfirm based on latency. 

We can do either or both slippage on the preconf tip or add some slack in the validation, so P2..Px would still get accepted with P1 price.

To avoid that the price that users gets is too outdated the entity that provides the price to users (or the users) needs to be aware of how much gas has been preconfirmed.

**Preconfirmations that use a lot of gas could underpay**
A preconf that consumes 170k gas should pay more than one that consumes a 17k gas. Currently, I'm not sure how this aligns with the existing wallet flow, put for a preconf req we need to know the gas that will be consumed.

Since price is calulated in steps, not as an integral, the 210k gas preconf would also pay less than 10x21k gas preconfs. While this discrepancy might not be significant, we need to be mindful of the oportunity cost for proposers.

Alternatively, users can just pay the worst preconf price and alway be included.

After this I plan to:
- Avoid hardcoded gas limit as it might change soon
- Start using pricing model in validation

### DIFF
--- a/bolt-sidecar/src/state/pricing.rs
+++ b/bolt-sidecar/src/state/pricing.rs
@@ -143,6 +143,9 @@ mod tests {
             pricing.calculate_min_priority_fee(incoming_gas, preconfirmed_gas).unwrap();
 
         // Expected fee: ~19 Gwei
+        // This will likely never happen, since you want to reserve some gas
+        // on top of the block for MEV, but enforcing this is not the responsibility
+        // of the pricing model.
         assert!(
             (min_fee_wei as f64 - 19_175_357_339.0).abs() < 1_000.0,
             "Expected ~19,175,357,339 Wei, got {} Wei",
--- a/bolt-sidecar/src/state/pricing.rs
+++ b/bolt-sidecar/src/state/pricing.rs
@@ -29,6 +29,12 @@ pub enum PricingError {
     },
 }
 
+impl Default for PreconfPricing {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl PreconfPricing {
     /// Initializes a new PreconfPricing with default parameters.
     pub fn new() -> Self {
--- a/bolt-sidecar/src/state/pricing.rs
+++ b/bolt-sidecar/src/state/pricing.rs
@@ -1,0 +1,188 @@
+/// Handles pricing calculations for preconfirmations
+#[derive(Debug)]
+pub struct PreconfPricing {
+    block_gas_limit: u64,
+    base_multiplier: f64,
+    gas_scalar: f64,
+}
+
+/// Errors that can occur during pricing calculations
+#[derive(Debug, thiserror::Error)]
+pub enum PricingError {
+    /// Preconfirmed gas exceeds the block limit
+    #[error("Preconfirmed gas {0} exceeds block limit {1}")]
+    ExceedsBlockLimit(u64, u64),
+    /// Insufficient remaining gas for the incoming transaction
+    #[error("Insufficient remaining gas: requested {requested}, available {available}")]
+    /// Insufficient remaining gas for the incoming transaction
+    InsufficientGas {
+        /// Gas requested by the incoming transaction
+        requested: u64,
+        /// Gas available in the block
+        available: u64,
+    },
+    /// Incoming gas is zero
+    #[error("Invalid gas limit: Incoming gas ({incoming_gas}) is zero")]
+    InvalidGasLimit {
+        /// Gas required by the incoming transaction
+        incoming_gas: u64,
+    },
+}
+
+impl PreconfPricing {
+    /// Initializes a new PreconfPricing with default parameters.
+    pub fn new() -> Self {
+        Self { block_gas_limit: 30_000_000, base_multiplier: 0.019, gas_scalar: 1.02e-6 }
+    }
+
+    /// Calculate the minimum priority fee for a preconfirmation based on
+    /// https://research.lido.fi/t/a-pricing-model-for-inclusion-preconfirmations/9136
+    ///
+    /// # Arguments
+    /// * `incoming_gas` - Gas required by the incoming transaction
+    /// * `preconfirmed_gas` - Total gas already preconfirmed
+    ///
+    /// # Returns
+    /// * `Ok(f64)` - The minimum priority fee in Wei
+    /// * `Err(PricingError)` - If the calculation cannot be performed
+    pub fn calculate_min_priority_fee(
+        &self,
+        incoming_gas: u64,
+        preconfirmed_gas: u64,
+    ) -> Result<u64, PricingError> {
+        // Check if preconfirmed gas exceeds block limit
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,197 @@
...
+impl Default for PreconfPricing {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PreconfPricing {
+    /// Initializes a new PreconfPricing with default parameters.
+    pub fn new() -> Self {
+        Self { block_gas_limit: 30_000_000, base_multiplier: 0.019, gas_scalar: 1.02e-6 }
```
estensen: Follow-up to remove hardcoded gas limit as it will likely change to 36M soon

merklefruit (approved):

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+        // T = Inclusion tip per gas
+        // 30M = Current gas limit (36M soon?)
+        let after_gas = remaining_gas - incoming_gas;
+
+        // Calculate numerator and denominator for the logarithm
+        let numerator = self.gas_scalar * (remaining_gas as f64) + 1.0;
+        let denominator = self.gas_scalar * (after_gas as f64) + 1.0;
+
+        // Calculate gas used
+        let expected_val = self.base_multiplier * (numerator / denominator).ln();
```
estensen, line 91: Consider using `U256` to avoid using floats since it support `ln`

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[error("Invalid gas limit: Incoming gas ({incoming_gas}) is zero")]
+    InvalidGasLimit {
+        /// Gas required by the incoming transaction
+        incoming_gas: u64,
+    },
+}
+
+impl Default for PreconfPricing {
+    fn default() -> Self {
+        Self::new(30_000_000)
```
thedevbirb, line 34: Can you gate this to a constant?

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+impl PreconfPricing {
+    /// Initializes a new PreconfPricing with default parameters.
+    pub fn new(block_gas_limit: u64) -> Self {
+        Self { block_gas_limit, base_multiplier: 0.019, gas_scalar: 1.02e-6 }
```
thedevbirb, line 41: Same here, can you add them in constants with a reference comment on top of each one?

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+        }
+
+        // Check if there is enough gas remaining in the block
+        let remaining_gas = self.block_gas_limit - preconfirmed_gas;
+        if incoming_gas > remaining_gas {
+            return Err(PricingError::InsufficientGas {
+                requested: incoming_gas,
+                available: remaining_gas,
+            });
+        }
```
thedevbirb, line 76: All these validation steps can be moved to a separate method. Wdyt?

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        // T(IG,UG) = 0.019 * ln(1.02⋅10^-6(30M-UG)+1 / 1.02⋅10^-6(30M-UG-IG)+1) / IG
```
thedevbirb, line 78: Reminder that these calculations can be performed directly using wei as unit of measure and not ether by multiplying the `base_muliplier` of `0.019` to `10**18`. Then, you directly work with either `u128` or `U256`. There is the `U256::log` function that should perform the natural logarithm of two 256 bit unsigned integers.

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+        let min_fee_wei =
+            pricing.calculate_min_priority_fee(incoming_gas, preconfirmed_gas).unwrap();
+
+        // Pricing model article expects fee of 0.61 Gwei
+        assert!(
+            (min_fee_wei as f64 - 613_499_092.0).abs() < 1_000.0,
+            "Expected ~613,499,092 Wei, got {} Wei",
+            min_fee_wei
+        );
+    }
```
thedevbirb, line 121: We should find out why we're getting a 0.02 gwei difference between this code and their model. Perhaps working in wei instead of ether can help

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+        // Test last preconfirmed transaction (21k gas, almost 30M preconfirmed)
+        let incoming_gas = 21_000;
+        let preconfirmed_gas = 30_000_000 - 21_000;
+        let min_fee_wei =
+            pricing.calculate_min_priority_fee(incoming_gas, preconfirmed_gas).unwrap();
+
+        // Expected fee: ~19 Gwei
+        // This will likely never happen, since you want to reserve some gas
+        // on top of the block for MEV, but enforcing this is not the responsibility
+        // of the pricing model.
```
thedevbirb, line 173: I'd drop these types of comments since this might not be the intention of every proposer.

bolt-sidecar/src/state/pricing.rs
```diff
@@ -0,0 +1,271 @@
...
+        assert!(
+            (min_fee_wei as f64 - 19_175_357_339.0).abs() < 1_000.0,
+            "Expected ~19,175,357,339 Wei, got {} Wei",
+            min_fee_wei
+        );
+    }
```
thedevbirb, line 179: I'd add a disclaimer that these checks are mainly for tracking down whether some changes in the algo are highly different compared to previous result, but these target values are not to be considered "correct" since they have been produced with this function in the beginning.

thedevbirb (changes requested):

thedevbirb:
Added some comments!

bolt-sidecar/src/state/pricing.rs
estensen: agreed, addressing in PR rn

bolt-sidecar/src/state/pricing.rs
estensen: Won't look into it now, but created an issue to address it later
https://github.com/chainbound/bolt/issues/596

PR #594, opened 2024-12-18
chore(bolt-sidecar): get block gas limit from config

estensen:
`PooledTransactionsElement` are rugged so need https://github.com/chainbound/bolt/pull/571 first

### DIFF
--- a/bolt-sidecar/src/config/chain.rs
+++ b/bolt-sidecar/src/config/chain.rs
@@ -20,6 +20,9 @@ pub const DEFAULT_COMMITMENT_DEADLINE_IN_MILLIS: u64 = 8_000;
 /// Default slot time duration in seconds.
 pub const DEFAULT_SLOT_TIME_IN_SECONDS: u64 = 12;
 
+/// Default gas limit for the sidecar.
+pub const DEFAULT_GAS_LIMIT: u64 = 30_000_000;
+
 /// The domain mask for signing application-builder messages.
 pub const APPLICATION_BUILDER_DOMAIN_MASK: [u8; 4] = [0, 0, 0, 1];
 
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/config/chain.rs
```diff
@@ -20,6 +20,9 @@ pub const DEFAULT_COMMITMENT_DEADLINE_IN_MILLIS: u64 = 8_000;
 /// Default slot time duration in seconds.
 pub const DEFAULT_SLOT_TIME_IN_SECONDS: u64 = 12;
 
+/// Default gas limit for the sidecar.
+pub const DEFAULT_GAS_LIMIT: u128 = 30_000_000;
```
thedevbirb: Using `u64` for gas limit is more than enough, even u16 can make it :D

thedevbirb (approved):
small nit

bolt-sidecar/src/config/chain.rs
```diff
@@ -20,6 +20,9 @@ pub const DEFAULT_COMMITMENT_DEADLINE_IN_MILLIS: u64 = 8_000;
 /// Default slot time duration in seconds.
 pub const DEFAULT_SLOT_TIME_IN_SECONDS: u64 = 12;
 
+/// Default gas limit for the sidecar.
+pub const DEFAULT_GAS_LIMIT: u64 = 30_000_000;
```
estensen, line 24: technically it would fit in `u32` but using `u64` since it's that's what Alloy does

https://github.com/alloy-rs/alloy/blob/2dee4da1413c326e2f94675b4e5139a6df897397/crates/consensus/src/transaction/pooled.rs#L357

PR #599, opened 2024-12-19
feat(bolt-sidecar): use pricing model to validate priority fee

estensen:
Implement [pricing model](https://research.lido.fi/t/a-pricing-model-for-inclusion-preconfirmations/9136#p-19482-a-model-for-cumulative-proposer-rewards-13) pioneered by Nethermind Research.

The pricing model defines a pricing curve to ensure proposers don't give up any profits when preconfirming a transaciton. Proposers would want profit for the extra work they do and for the "risk free" rate they could get on the 1 ETH collateral they put down with liquid staking. We're currently researching what this profit should be, but for now we default it to 2 GWEI but make it configurable.

In many cases the proposer will outsource the pricing to the Firewall, but this will PR will make sure proposers don't accept a preconf where they lose money.

### DIFF
--- a/bolt-sidecar/.env.example
+++ b/bolt-sidecar/.env.example
@@ -33,8 +33,8 @@ BOLT_SIDECAR_BUILDER_PRIVATE_KEY=
 BOLT_SIDECAR_MAX_COMMITMENTS_PER_SLOT=128
 # Max committed gas per slot
 BOLT_SIDECAR_MAX_COMMITTED_GAS_PER_SLOT=10_000_000
-# Min priority fee to accept for a commitment
-BOLT_SIDECAR_MIN_PRIORITY_FEE=2000000000 # 2 Gwei = 2 * 10^9 wei
+# Min profit per gas to accept a commitment
+BOLT_SIDECAR_MIN_PROFIT=2000000000 # 2 Gwei = 2 * 10^9 wei
 
 # Chain configuration
 # Chain on which the sidecar is running
--- a/bolt-sidecar/src/primitives/commitment.rs
+++ b/bolt-sidecar/src/primitives/commitment.rs
@@ -191,7 +191,7 @@ impl InclusionRequest {
             let min_priority_fee =
                 pricing.calculate_min_priority_fee(tx.gas_limit(), preconfirmed_gas)?;
 
-            let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or(0);
+            let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or_default();
             if tip < min_priority_fee as u128 {
                 return Ok(false);
             }
--- a/bolt-sidecar/src/primitives/commitment.rs
+++ b/bolt-sidecar/src/primitives/commitment.rs
@@ -7,7 +7,10 @@ use alloy::{
 };
 use serde::{de, Deserialize, Deserializer, Serialize};
 
-use crate::crypto::SignerECDSA;
+use crate::{
+    crypto::SignerECDSA,
+    state::{pricing::PricingError, PreconfPricing},
+};
 
 use super::{deserialize_txs, serialize_txs, FullTransaction, TransactionExt};
 
@@ -173,10 +176,29 @@ impl InclusionRequest {
     /// Validates the priority fee against a minimum priority fee.
     /// Returns `true` if the "effective priority fee" is greater than or equal to the set minimum
     /// priority fee, `false` otherwise.
-    pub fn validate_min_priority_fee(&self, max_base_fee: u128, min_priority_fee: u128) -> bool {
-        self.txs.iter().all(|tx| {
-            tx.effective_tip_per_gas(max_base_fee).map_or(false, |tip| tip >= min_priority_fee)
-        })
+    /// Returns an error if min priority fee cannot be calculated.
+    pub fn validate_min_priority_fee(
+        &self,
+        pricing: &PreconfPricing,
+        preconfirmed_gas: u64,
+        max_base_fee: u128,
+    ) -> Result<bool, PricingError> {
+        // Each included tx will move the price up
+        // So we need to calculate the minimum priority fee for each tx
+        let mut local_preconfirmed_gas = preconfirmed_gas;
+        for tx in &self.txs {
+            // Calculate minimum required priority fee for this transaction
+            let min_priority_fee =
+                pricing.calculate_min_priority_fee(tx.gas_limit(), preconfirmed_gas)?;
+
+            let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or(0);
+            if tip < min_priority_fee as u128 {
+                return Ok(false);
+            }
+            // Increment the preconfirmed gas for the next transaction in the bundle
+            local_preconfirmed_gas = local_preconfirmed_gas.saturating_add(tx.gas_limit());
+        }
+        Ok(true)
     }
 
     /// Returns the total gas limit of all transactions in this request.
--- a/bolt-sidecar/src/primitives/commitment.rs
+++ b/bolt-sidecar/src/primitives/commitment.rs
@@ -181,6 +181,7 @@ impl InclusionRequest {
         &self,
         pricing: &PreconfPricing,
         preconfirmed_gas: u64,
+        min_inclusion_profit: u64,
         max_base_fee: u128,
     ) -> Result<bool, PricingError> {
         // Each included tx will move the price up
@@ -188,8 +189,9 @@ impl InclusionRequest {
         let mut local_preconfirmed_gas = preconfirmed_gas;
         for tx in &self.txs {
             // Calculate minimum required priority fee for this transaction
-            let min_priority_fee =
-                pricing.calculate_min_priority_fee(tx.gas_limit(), preconfirmed_gas)?;
+            let min_priority_fee = pricing
+                .calculate_min_priority_fee(tx.gas_limit(), preconfirmed_gas)?
+                + min_inclusion_profit;
 
             let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or_default();
             if tip < min_priority_fee as u128 {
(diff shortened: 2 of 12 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -123,6 +124,12 @@ impl ValidationError {
     }
 }
 
+impl From<pricing::PricingError> for ValidationError {
+    fn from(err: pricing::PricingError) -> Self {
+        ValidationError::Internal(format!("Pricing error: {}", err))
+    }
+}
```
estensen: I think some validation in `validate_request()` is duplicated in `calculate_min_priority_fee()` so we might not need this after some deduplication

bolt-sidecar/src/state/execution.rs
```diff
@@ -311,8 +318,11 @@ impl<C: StateFetcher> ExecutionState<C> {
             return Err(ValidationError::BaseFeeTooLow(max_basefee));
         }
 
-        // Ensure max_priority_fee_per_gas is greater than or equal to min_priority_fee
-        if !req.validate_min_priority_fee(max_basefee, self.limits.min_priority_fee) {
+        // Create pricing calculator
+        let pricing = pricing::PreconfPricing::new(self.validation_params.block_gas_limit);
```
merklefruit: I think the pricing calculator could be part of the execution state itself instead of being recreated every time

bolt-sidecar/src/state/execution.rs
```diff
@@ -123,6 +124,12 @@ impl ValidationError {
     }
 }
 
+impl From<pricing::PricingError> for ValidationError {
+    fn from(err: pricing::PricingError) -> Self {
+        Self::Internal(format!("Pricing error: {}", err))
+    }
+}
```
merklefruit: this could also be done above as 

```rust
#[derive(Debug, Error)]
pub enum ValidationError {
	#[error("Pricing error: {0}")
	Pricing(#[from] pricing::PricingError)
```

returning "internal error" also means that bolt-rpc won't get feedback on why the commitment failed, which in my opinion we should provide for pricing errors. If this was intentional then it's also fine but let's discuss

merklefruit (approved):
Great!

bolt-sidecar/src/state/execution.rs
estensen: agreed!

bolt-sidecar/src/state/execution.rs
estensen: done

bolt-sidecar/src/primitives/commitment.rs
```diff
@@ -173,10 +176,29 @@ impl InclusionRequest {
...
+        preconfirmed_gas: u64,
+        max_base_fee: u128,
+    ) -> Result<bool, PricingError> {
+        // Each included tx will move the price up
+        // So we need to calculate the minimum priority fee for each tx
+        let mut local_preconfirmed_gas = preconfirmed_gas;
+        for tx in &self.txs {
+            // Calculate minimum required priority fee for this transaction
+            let min_priority_fee =
+                pricing.calculate_min_priority_fee(tx.gas_limit(), preconfirmed_gas)?;
```
thedevbirb: Can this function return  an `u128`?

bolt-sidecar/src/primitives/commitment.rs
```diff
@@ -173,10 +176,29 @@ impl InclusionRequest {
...
+
+            let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or(0);
```
thedevbirb:
```suggestion
            let tip = tx.effective_tip_per_gas(max_base_fee).unwrap_or_default();
```

thedevbirb (approved):

bolt-sidecar/src/primitives/commitment.rs
estensen: Yes it could, but why is it necessary?
For 30M gas the expected result is [613_499_092..19_175_357_339]

bolt-sidecar/src/primitives/commitment.rs
estensen: ty for the nit

bolt-sidecar/src/primitives/commitment.rs
thedevbirb: Because on the line below you do a conversion to an u128, but it's not a strong requirement tho

bolt-sidecar/.env.example
```diff
@@ -33,8 +33,8 @@ BOLT_SIDECAR_BUILDER_PRIVATE_KEY=
 BOLT_SIDECAR_MAX_COMMITMENTS_PER_SLOT=128
 # Max committed gas per slot
 BOLT_SIDECAR_MAX_COMMITTED_GAS_PER_SLOT=10_000_000
-# Min priority fee to accept for a commitment
-BOLT_SIDECAR_MIN_PRIORITY_FEE=2000000000 # 2 Gwei = 2 * 10^9 wei
+# Min profit per gas to accept a commitment
+BOLT_SIDECAR_MIN_PROFIT=2000000000 # 2 Gwei = 2 * 10^9 wei
```
estensen, line 37: This needs to be calculated and based on profit a proposer needs to exceed the "risk free" return of liquid staking on the collateral

bolt-sidecar/src/primitives/commitment.rs
```diff
@@ -173,10 +176,31 @@ impl InclusionRequest {
     /// Validates the priority fee against a minimum priority fee.
     /// Returns `true` if the "effective priority fee" is greater than or equal to the set minimum
     /// priority fee, `false` otherwise.
-    pub fn validate_min_priority_fee(&self, max_base_fee: u128, min_priority_fee: u128) -> bool {
-        self.txs.iter().all(|tx| {
-            tx.effective_tip_per_gas(max_base_fee).map_or(false, |tip| tip >= min_priority_fee)
-        })
+    /// Returns an error if min priority fee cannot be calculated.
+    pub fn validate_min_priority_fee(
```
mempirate, line 180: I would move the bulk logic of this function to `ExecutionState`, and just have this validate that the `effective_tip_per_gas` is greater than the result of running the pricing algo. So the function can actually remain the same as before.

bolt-sidecar/src/state/execution.rs
```diff
@@ -311,8 +320,13 @@ impl<C: StateFetcher> ExecutionState<C> {
...
-        // Ensure max_priority_fee_per_gas is greater than or equal to min_priority_fee
-        if !req.validate_min_priority_fee(max_basefee, self.limits.min_priority_fee) {
+        // Ensure max_priority_fee_per_gas is greater than or equal to the calculated min_priority_fee
+        if !req.validate_min_priority_fee(
+            &self.pricing,
+            template_committed_gas,
+            self.limits.min_inclusion_profit,
+            max_basefee,
+        )? {
             return Err(ValidationError::MaxPriorityFeePerGasTooLow);
```
mempirate, line 330: We should return the value of the minimum fee in this error, so it can be propagated to any consumers

testnets/holesky/README.md
```diff
@@ -777,11 +777,11 @@ Options:
           [env: BOLT_SIDECAR_MAX_COMMITTED_GAS=]
           [default: 10000000]
 
-      --min-priority-fee <MIN_PRIORITY_FEE>
-          Min priority fee to accept for a commitment
+      --min-inclusion_profit <MIN_INCLUSION_PROFIT>
```
mempirate: Should be `--min-inclusion-profit` (`-` instead of `_`)

bolt-sidecar/src/state/pricing.rs
```diff
@@ -195,6 +185,55 @@ mod tests {
...
+            small_fee_sum += small_fee;
+            preconfirmed_gas_small += small_gas;
+        }
+
+        // Moving on the pricing curve in 10 steps should cost
+        // the same as moving in one big step per gas.
+        let small_sum_fee_avg = small_fee_sum / 10;
+
+        assert!(
+            (big_fee as f64 - small_sum_fee_avg as f64).abs() < 1_000.0,
```
mempirate, line 231: Did we figure this out?

bolt-sidecar/src/state/execution.rs
```diff
@@ -222,6 +230,7 @@ impl<C: StateFetcher> ExecutionState<C> {
             kzg_settings: EnvKzgSettings::default(),
             // TODO: add a way to configure these values from CLI
             validation_params: ValidationParams::new(gas_limit),
+            pricing: PreconfPricing::new(gas_limit),
```
mempirate: Nit but can we rename this to `pricer: InclusionPricer`? Less ambiguous if we have different preconf types in the future.

mempirate (changes requested):

bolt-sidecar/.env.example
estensen: That assumtion was wrong since we're using liquid restaking.

bolt-sidecar/src/state/pricing.rs
estensen: I think it was a misunderstanding on my part.
The current test validates that one 210k gas tx gives the same avg gas price as 10x21k gas txs.

bolt-sidecar/src/state/execution.rs
estensen: Is it ok if I create an issue and follow up in a separate PR? Keen on getting this merged.

bolt-sidecar/src/primitives/commitment.rs
estensen: Is it ok if I create an issue and follow up in a separate PR? Keen on getting this merged.

bolt-sidecar/src/state/execution.rs
estensen: https://github.com/chainbound/bolt/issues/647

mempirate (approved):
Good stuff

bolt-sidecar/src/primitives/commitment.rs
estensen: https://github.com/chainbound/bolt/issues/648

bolt-sidecar/src/primitives/commitment.rs
estensen: might do in follow-up

PR #620, opened 2024-12-27
Faheelsattar/remove expiry flag

### LINKED ISSUE #585: Remove `--expiry` flag when registering operator with EigenLayer
We need the `expiry` field and the `salt` to be compliant with the EigenLayer contracts. In particular the command `bolt eigenlayer register` will call the contract `AVSDirectory.registerOperatorToAVS` (ref: https://github.com/Layr-Labs/eigenlayer-contracts/blob/testnet-holesky/src/contracts/core/AVSDirectory.sol#L64-L71). The expiry, along with the salt, are mainly security checks for avoiding the same signature to be used many times.

Since this command will produce the signature and use it immediately, we can avoid asking the user such request and produce automatically a very short timestamp expiry. The docs should be changed accordingly.

faheelsattar:
Closes #585

### DIFF
--- a/bolt-cli/src/commands/operators/eigenlayer.rs
+++ b/bolt-cli/src/commands/operators/eigenlayer.rs
@@ -118,14 +118,14 @@ impl EigenLayerSubcommand {
                     AVSDirectory::new(deployments.eigen_layer.avs_directory, provider);
                 
                 const EXPIRY_DURATION :TimeDelta = Duration::minutes(20);
-                let expiry_time = Utc::now() + EXPIRY_DURATION;
-
+                let expiry = U256::from((Utc::now() + EXPIRY_DURATION).timestamp());
+                
                 let signature_digest_hash = avs_directory
                     .calculateOperatorAVSRegistrationDigestHash(
                         signer.address(),
                         bolt_avs_address,
                         salt,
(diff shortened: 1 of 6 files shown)


bolt-cli/src/commands/operators/eigenlayer.rs
```diff
@@ -113,20 +116,28 @@ impl EigenLayerSubcommand {
 
                 let avs_directory =
                     AVSDirectory::new(deployments.eigen_layer.avs_directory, provider);
+                
+                const EXPIRY_DURATION :TimeDelta = Duration::minutes(20);
+                let expiry_time = Utc::now() + EXPIRY_DURATION;
```
merklefruit: nit: 
```suggestion
                let expiry = U256::from((Utc::now() + EXPIRY_DURATION).timestamp());
```

merklefruit (changes requested):
Wonderful, thanks for the contribution! I left a small nit but other than that LGTM!

bolt-cli/src/commands/operators/eigenlayer.rs
faheelsattar: added the suggestion

merklefruit (approved):

PR #632, opened 2025-01-06
Genericise preconfirmation pricing

jmcph4:
# Description

Pricing preconfirmations is an obviously important aspect of any successful implementation yet their treatment is still extremely limited (with the existing Lido research being the only publicly disclosed method that the author is aware of). This is largely due to the preconfirmation space itself being so nascent. This will surely change though, and this PR aims to make the addition of new pricing models to the Bolt sidecar easier.

# Issue

N/A

# Changes

 - Define `Pricing` trait
 - Improve/fix documentation comments, where appropriate
 - Move existing Lido model implementation to its own submodule

# Rationale

Paves the way for future extensions to preconfirmation pricing capabilities; namely the addition of alternative pricing models

### DIFF
--- a/bolt-sidecar/src/state/mod.rs
+++ b/bolt-sidecar/src/state/mod.rs
@@ -13,7 +13,7 @@ pub use execution::{ExecutionState, ValidationError};
 
 /// Module to calculate pricing.
 pub mod pricing;
-pub use pricing::PreconfPricing;
+pub use pricing::LidoPricing;
 
 /// Module to fetch state from the Execution layer.
 pub mod fetcher;
--- a/bolt-sidecar/src/state/pricing/lido.rs
+++ b/bolt-sidecar/src/state/pricing/lido.rs
@@ -1,65 +1,37 @@
-/// Gas limit constants
-pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 30_000_000;
+//! Implementation of preconfirmation pricing as presented in
+//! <https://research.lido.fi/t/a-pricing-model-for-inclusion-preconfirmations/9136>.
+use super::{Pricing, PricingError, DEFAULT_BLOCK_GAS_LIMIT};
 
 /// Fee calculation constants from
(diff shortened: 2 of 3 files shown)


merklefruit:
Thanks for the contribution @jmcph4!

Just wanted to add some more context regarding the pricing that might be unclear from looking at the code here, and in general point to a direction for further updates:

As we currently stand, preconf pricing logic can occur in two places:

1. bolt-sidecar, if a node operator wishes to run everything locally
2. bolt-rpc, if a node operator wishes to set their `operator_url` endpoint to a third party RPC.

Why would a proposer want to delegate their `operator_url` to a third party? A few reasons:

1. Privacy & DDoS-protection, because broadcasting their own operator_url means that anyone would be aware of their IP. Instead of specifying their own IP, they can use a "firewall RPC" endpoint that will relay requests to them.
2. Better pricing: trusting a third-party RPC for pricing means that this complexity can be pushed away from proposers, while still leaving them the full rights of accepting or denying the requests if they wish.

With this in mind, the pricing logic on the sidecar will really only be used if the proposers wish to not use the firewall RPC feature. And in this case, we feel like having widely custom pricing implementations based on proposers choice is not the best approach as it increases complexity a lot in this phase.

We want to experiment and encourage builders to come up with more precise pricing strategies, but at the same time this comes at cost of complexity and we'd like to keep it simple for now.

In terms of this specific PR, the biggest issue for me is that if we were to make a generic pricing trait it would probably not take the same inputs as it does now: `incoming_gas` and `preconfirmed_gas` are just some of the variables one could use to determine a pricing strategy (for instance, you may want to price based on calldata content/length, or destination address, or something else entirely!)

Given these points, I think the best move forward is to close this for now and reopen it when we have a clearer idea on how proposers actually want to communicate their pricing preferences on a live network

Hopefully this all makes sense! Please don't see it as a discouragement to contribute to bolt, and instead as a way to understand the context a little bit better. We appreciate this a lot! 🫡

PR #644, opened 2025-01-07
feat(cli): add command to output a compressed pubkey hash

merklefruit:
This PR adds a "pubkey-hash" command to the bolt CLI, which can be useful for debugging purposes. 
It simply requires a pubkey as argument. Here is an example:

```sh
bolt pubkey-hash 0x90b7beb2568094bec116f264b63d259c9205e5a969e11316ec57b57ef827c206504730ddbfccf7731a802cb99401446a

0x43068754c6062ac20e198e99f87d492973592ccd
```

this command is hidden from clap so it's not shown to users in the help menu.

### DIFF
--- a/bolt-cli/src/cli.rs
+++ b/bolt-cli/src/cli.rs
@@ -44,6 +44,10 @@ pub enum Cmd {
 
     /// Useful data generation commands.
     Generate(GenerateCommand),
+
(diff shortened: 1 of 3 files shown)


mempirate (approved):
Great, very useful in debugging indeed. Also didn't know about hiding clap commands!

PR #649, opened 2025-01-08
feat(bolt-sidecar): add version to metadata endpoint

estensen:
For each lookahead window the Firewall will query proposers to get their version and profit expectation.
Currently `bolt_getVersion` is queried, but we wanna consolidate so only one API call is needed to get all the metadata info.

Explicitly check each field in test to catch regression. Current assert will not catch changes to `LimitsOpts` since we compare with default.

Discovered that https://github.com/chainbound/bolt/pull/599 broke `bolt_metadata` but I'm not goint to revert and bump the API since it's not in use yet.

Next PR will add profit to the `bolt_metadata` endpoint.

### DIFF
--- a/bolt-sidecar/src/api/commitments/server/handlers.rs
+++ b/bolt-sidecar/src/api/commitments/server/handlers.rs
@@ -1,4 +1,4 @@
-use std::{num::NonZero, sync::Arc};
+use std::sync::Arc;
 
 use axum::{
     body::Body,
@@ -35,28 +35,11 @@ use super::CommitmentsApiInner;
 /// limits configuration with version information in a flat structure
 #[derive(Debug, Serialize, Deserialize)]
 pub struct MetadataResponse {
-    /// Max number of commitments to accept per slot
-    pub max_commitments_per_slot: NonZero<usize>,
-    /// Max committed gas per slot
-    pub max_committed_gas_per_slot: NonZero<u64>,
-    /// Min priority fee to accept for a commitment
-    pub min_priority_fee: u128,
-    /// The maximum size in MiB of the account states
-    pub max_account_states_size: NonZero<usize>,
+    /// The operational limits of the sidecar
+    #[serde(flatten)]
+    pub limits: LimitsOpts,
     /// The version of the Bolt sidecar
     pub version: String,
-}
-
-impl From<(LimitsOpts, String)> for MetadataResponse {
-    fn from((limits, version): (LimitsOpts, String)) -> Self {
-        Self {
-            max_commitments_per_slot: limits.max_commitments_per_slot,
-            max_committed_gas_per_slot: limits.max_committed_gas_per_slot,
-            min_priority_fee: limits.min_priority_fee,
-            max_account_states_size: limits.max_account_states_size,
-            version,
-        }
-    }
 }
 
 /// Handler function for the root JSON-RPC path.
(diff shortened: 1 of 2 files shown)


bolt-sidecar/src/api/commitments/handlers.rs
```diff
@@ -26,6 +28,34 @@ use super::{
...
+    pub max_commitments_per_slot: NonZero<usize>,
+    /// Max committed gas per slot
+    pub max_committed_gas_per_slot: NonZero<u64>,
+    /// Min priority fee to accept for a commitment
+    pub min_priority_fee: u128,
+    /// The maximum size in MiB of the account states
+    pub max_account_states_size: NonZero<usize>,
+    /// The version of the Bolt sidecar
+    pub version: String,
+}
```
estensen: Tried to use flatten instead, but had problems deserializing the response in the test because `serde_json` doesn't support `u128`

```rust
pub struct MetadataResponse<T> {
    /// The operational limits of the sidecar
    #[serde(flatten)]
    pub limits: LimitOp,
    /// The version of the Bolt sidecar
    pub version: String,
}
```

estensen:
Tests are green locally

estensen:
ah, PR is probably running as if it was merged with `unstable` need rebase

bolt-sidecar/src/api/commitments/handlers.rs
thedevbirb: I don't remember, is it necessary to use `u128` tho? If in wei, a `u64` can express a `min_priority_fee` of `(2**64 - 1) / 10**18 == 18 ETH`, should be enough :D

bolt-sidecar/src/api/commitments/handlers.rs
```diff
@@ -43,9 +73,12 @@ pub async fn rpc_entrypoint(
         })),
 
         GET_METADATA_METHOD => {
+            let metadata: MetadataResponse =
+                (api.limits(), BOLT_SIDECAR_VERSION.to_string()).into();
```
thedevbirb: This doesn't feel very idiomatic, does the following work?
```rs
let metadata = MetadataResponse {
   version: BOLT_SIDECAR_VERSION.to_string(),
   ..api.limits()
}
```

bolt-sidecar/src/api/commitments/handlers.rs
```diff
@@ -43,9 +73,12 @@ pub async fn rpc_entrypoint(
...
+
             let response = JsonResponse {
                 id: payload.id,
-                result: serde_json::to_value(api.limits()).expect("infallible"),
+                result: serde_json::to_value(metadata).expect("infallible"), // metadata only contains primitive types
```
thedevbirb: I think this message is more useful inside the `expect` itself!

bolt-sidecar/src/api/commitments/handlers.rs
estensen: No, because `api.limits()` returns `LimitOpts` and not `MetadataResponse`

estensen:
Since `min_inclusion_profit` is `u64` I could revert to do flatten

thedevbirb (approved):

bolt-sidecar/src/api/commitments/handlers.rs
thedevbirb: ah right, mb

bolt-sidecar/src/api/commitments/handlers.rs
merklefruit: oof, I don't love that tbh! Why put this limit in the code? let's either find a type that works like alloy::primitives::U128 or what @estensen did IMO

PR #653, opened 2025-01-10
chore(bolt-sidecar): remove BOLT_SIDECAR_MAX_COMMITMENTS_PER_SLOT

estensen:
We think it's sufficient for the proposers to be able to configure `max_committed_gas_per_slot` and `min_inclusion_profit`. Removing `max_committed_gas_per_slot` will make it remove routing complexity from the Firewall.

We can always add it back later if we think we need it.


merklefruit (approved):
lgtm

thedevbirb (approved):
great

PR #655, opened 2025-01-10
Feature flag to exclude high compute `test_keystore_signer` tests

### LINKED ISSUE #344: Investigate why `test_keystore_signer` takes so long
Reference: https://github.com/chainbound/bolt/blob/7f7671840ab8d63f734bb0dc19cb8d1d2c91df06/bolt-sidecar/src/signer/keystore.rs#L213-L371 
Currently takes between 1-2mins alone.

If this can be reduced that's great, otherwise let's document it well.

0ex-d:
Decoding the keystore would take seconds in reality because of the high iteration count required by <a href="https://eips.ethereum.org/EIPS/eip-2335#test-cases" target="_blank">EIP-2335</a> test vectors. 

### Improvements:
- Since tests are run periodically, we can rather flag with a feature (`keystore-tests`) only to be run explicitly on CI/CD workflows.
- Cleanup `test-voting-keystore.json` after test.

For more information, please refer to this [issue](https://github.com/chainbound/bolt/issues/344).
Closes #344


merklefruit (approved):
This is awesome, thanks a lot! 🚀

thedevbirb:
Hey @0ex-d, thanks for picking up this old issue, I almost forgot about it :). 
You're right that the test takes long because of the high iterations count set by the param `n == 2**18 == 262144` provided to the key derivation functions.

The PR is good! I like the idea of the feature flag. I just have a small nit/question: why did you choose to move the unit test in a separate module?

estensen (approved):

0ex-d:
Thanks guys for the feedback and for confirming the PR!

First obvious solution was to just do:
```rust
#[cfg(feature = "keystore-tests")]
#[test]
    fn test_keystore_signer() {}
```
Rust **clippy/linter** would render these lines as `warnings` therefore breaking UI:
https://github.com/chainbound/bolt/blob/87873739a524eebf58eeda55b2168f0417fba5a6/bolt-sidecar/src/signer/keystore.rs#L190-L213

Well, another fix would be to wrap most of the lines with `#[cfg(feature = "keystore-tests")]`:
```rust
#[cfg(test)]
mod tests {
    #[cfg(feature = "keystore-tests")]
    use std::{
        fs::File,
        io::Write,
        path::{Path, PathBuf},
    };
    #[cfg(feature = "keystore-tests")]
    use blst::min_pk::SecretKey;
...
```

So the tradeoff was keeping the test mod in the same file (plus loads of config macros) or just wrap them all within a single file without breaking linter/UI which is often uncommon to seperate test concerns.

Let me know if you'd prefer a different structure!

thedevbirb:
Thanks for the explanation! But since you added:
```rs
#[cfg(feature = "keystore-tests")]
#[cfg(test)]
mod tests {
	/* code */
}
```
to `keystore_tests.rs`, couldn't you do to the same for `keystore.rs`? Or am I missing something?

0ex-d:
Thats a very good point! pushing a change now.

thedevbirb (approved):
Thanks!

PR #656, opened 2025-01-10
fix(bolt-sidecar): correctly get preconf gas limit from env

estensen:
Currently we get `BOLT_SIDECAR_GAS_LIMIT` which is the gas limit for a single tx. It defaults to `DEFAULT_GAS_LIMIT` which is 30M. That means pricing input validation will not fail when trying to use more gas than what's left for preconfs in a block. Other valdiation might reject this earlier, but I'd like it to be rejected here as well.

I think it's confusing to have to ENV vars that do the same, so removed `BOLT_SIDECAR_GAS_LIMIT` and replace it with `BOLT_SIDECAR_MAX_COMMITTED_GAS`

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -179,6 +179,7 @@ pub struct ExecutionState<C> {
 /// Other values used for validation.
 #[derive(Debug)]
 pub struct ValidationParams {
+    // The maximum amount of gas reserved to preconfirmations for blocks.
     pub preconf_gas_limit: u64,
     pub max_tx_input_bytes: usize,
     pub max_init_code_byte_size: usize,
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -230,7 +230,7 @@ impl<C: StateFetcher> ExecutionState<C> {
             kzg_settings: EnvKzgSettings::default(),
             // TODO: add a way to configure these values from CLI
             validation_params: ValidationParams::new(gas_limit),
-            pricing: InclusionPricer::new(gas_limit),
+            pricing: InclusionPricer::new(limits.max_committed_gas_per_slot.get()),
         })
     }
 
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -179,7 +179,7 @@ pub struct ExecutionState<C> {
 /// Other values used for validation.
 #[derive(Debug)]
 pub struct ValidationParams {
-    pub block_gas_limit: u64,
+    pub preconf_gas_limit: u64,
     pub max_tx_input_bytes: usize,
     pub max_init_code_byte_size: usize,
 }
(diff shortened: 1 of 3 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -229,8 +225,8 @@ impl<C: StateFetcher> ExecutionState<C> {
             // Load the default KZG settings
             kzg_settings: EnvKzgSettings::default(),
             // TODO: add a way to configure these values from CLI
```
thedevbirb, line 228: Is this comment still relevant?

bolt-sidecar/src/state/execution.rs
```diff
@@ -286,7 +282,7 @@ impl<C: StateFetcher> ExecutionState<C> {
             return Err(ValidationError::TransactionSizeTooHigh);
         }
 
-        // Check if the gas limit is higher than the maximum block gas limit
+        // Check if the gas limit is higher than the preconf block gas limit
         if req.gas_limit() > self.validation_params.block_gas_limit {
```
thedevbirb: Do you think renaming the property `block_gas_limit` of `validation_params` to `preconf_gas_limit` enhances clarity?

thedevbirb (approved):
Good catch! A couple of minor questions

bolt-sidecar/src/state/execution.rs
estensen: yes, currently only `preconf_gas_limit` is configurable

```rust
/// Other values used for validation.
#[derive(Debug)]
pub struct ValidationParams {
    pub preconf_gas_limit: u64,
    pub max_tx_input_bytes: usize,
    pub max_init_code_byte_size: usize,
}

impl ValidationParams {
    pub fn new(gas_limit: u64) -> Self {
        Self {
            preconf_gas_limit: gas_limit,
            max_tx_input_bytes: 4 * 32 * 1024,
            max_init_code_byte_size: 2 * 24576,
        }
    }
}
```

Haven't been able to review the usefulnes of the others yet, so will leave the TODO for now

bolt-sidecar/src/state/execution.rs
estensen: yes, good idea
https://github.com/chainbound/bolt/pull/656/commits/9cd5b9c026acd66a624709e502c9df6176547136

bolt-sidecar/src/state/execution.rs
```diff
@@ -179,15 +179,15 @@ pub struct ExecutionState<C> {
 /// Other values used for validation.
 #[derive(Debug)]
 pub struct ValidationParams {
-    pub block_gas_limit: u64,
+    pub preconf_gas_limit: u64,
```
thedevbirb, line 183:
```suggestion
    // The maximum amount of gas reserved to preconfirmations for blocks.
    pub preconf_gas_limit: u64,
```

thedevbirb (approved):
Last nit

bolt-sidecar/src/state/execution.rs
mempirate: Sorry about nitpicking the naming here, but can we change this to `inclusion_gas_limit`? If introduce new commitment types, or execution preconfs, the naming is less ambiguous. In general let's try to use inclusion commitment or something similar instead of preconf in the codebase

estensen: nitpicking is good 👑👑

merklefruit (approved):
lgtm!

PR #657, opened 2025-01-10
chore(docker/holesky): cleanup docker compose files

thedevbirb:
This PR tries to align some diffs we have between the code on GitHub and our setup on remotebeast, so that we don't have conflicts and headaches when pulling new changes.

Changelog:
- builder -- bump image; on remotebeast it is running a newer one
- helix -- on remotebeast we're running also the website service, which is missing on github
- networks -- unify networks on a single `bolt-holesky-default`
- format run-builder and run-bn


testnets/holesky/docker-compose.yml
```diff
@@ -82,7 +76,7 @@ volumes:
     driver: local
 
 networks:
-  bolt-default:
+  bolt-holesky-default:
     driver: bridge
```
mempirate: You will have to add `name: bolt-holesky-default` here. Otherwise docker compose will prefix it with some bs and the name will differ in `docker-compose.pbs.yml`

thedevbirb: Updated this to drop all references to custom networks, after internal discussion we will use `host.docker.internal` as DNS and adding:
```yaml
    extra_hosts:
      - host.docker.internal:host-gateway
```
On every service that needs to communicate with others which expose a port on the host machine

testnets/holesky/docker-compose.yml
```diff
@@ -71,21 +66,10 @@ services:
       - ./grafana/datasources:/etc/grafana/provisioning/datasources
       - grafana-data:/var/lib/grafana
     networks:
-      - bolt-default
-    depends_on:
-      - bolt-prometheus-holesky
+      - bolt-holesky-default
```
merklefruit: what is this network? i thought you said we don't use this anymore

merklefruit (changes requested):
Looks good, just one small q!

mempirate (approved):
LGTM, let's merge when tested

merklefruit (approved):

PR #674, opened 2025-01-13
Chore: Bolt-CLI: Use library for confirmation logic

### LINKED ISSUE #406: Use library for confirmation logic
In Bolt-CLI we use a little function to ask user confirmations, but there is library with other features like [docs.rs/inquire/latest/inquire/struct.Confirm.html](https://docs.rs/inquire/latest/inquire/struct.Confirm.html).

Reference discussion: https://github.com/chainbound/bolt/pull/404#discussion_r1851618402

0ex-d:
Bolt-cli now uses [inquire crate](https://docs.rs/inquire/latest/inquire/struct.Confirm.html).
### Improvements:
Impl `request_confirmation()` to `Cmd` struct.
For more information, please refer to this [issue](https://github.com/chainbound/bolt/issues/406).
Closes #406

### DIFF
--- a/bolt-cli/src/cli.rs
+++ b/bolt-cli/src/cli.rs
@@ -15,7 +15,6 @@ use crate::{
     common::{keystore::DEFAULT_KEYSTORE_PASSWORD, parse_ether_value},
     contracts::EigenLayerStrategy,
 };
-use tracing::info;
 
 /// `bolt` is a CLI tool to interact with bolt Protocol ✨
 #[derive(Parser, Debug, Clone)]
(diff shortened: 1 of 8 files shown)


bolt-cli/src/cli.rs
```diff
@@ -63,6 +64,28 @@ impl Cmd {
...
+                    return;
+                }
+                info!("Aborting");
+                std::process::exit(0);
+            })
+            .unwrap_or_else(|err| {
+                info!("confirmation exited: {}", err);
+                std::process::exit(0);
+            })
+    }
```
merklefruit: this is awesome, however could you move this function back in bolt-cli/src/common/mod.rs? 
We usually avoid using static methods. 

Thanks for the contribution!

0ex-d: Great observation!

merklefruit (approved):
LGTM

bolt-cli/src/common/mod.rs
```diff
@@ -105,3 +106,25 @@ pub fn try_parse_contract_error<T: SolInterface>(error: ContractError) -> Result
...
+            if proceed {
+                return;
+            }
+            info!("Aborting");
+            std::process::exit(0);
+        })
+        .unwrap_or_else(|err| {
+            info!("confirmation exited: {}", err);
+            std::process::exit(0);
+        })
```
thedevbirb, line 129: Can you do some pattern matching based on `err`? Reference: https://docs.rs/inquire/latest/inquire/error/enum.InquireError.html

For example upon `OperationCanceled` or `OperatorInterrupted` we should log `"Aborting"` as well, otherwise log with `error` directive and return different error codes

thedevbirb (changes requested):
Thanks for the contribution! A minor request

PR #680, opened 2025-01-15
chore(bolt-sidecar): log unexpected websocket messages

estensen:
Help detect protocol misuse

### DIFF
--- a/bolt-sidecar/src/api/commitments/firewall/processor.rs
+++ b/bolt-sidecar/src/api/commitments/firewall/processor.rs
@@ -17,7 +17,7 @@ use tokio_tungstenite::{
     tungstenite::{self, Message},
     MaybeTlsStream, WebSocketStream,
 };
-use tracing::{error, info, trace, warn};
+use tracing::{debug, error, info, trace, warn};
 use uuid::Uuid;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


bolt-sidecar/src/api/commitments/firewall/processor.rs
```diff
@@ -222,7 +222,22 @@ impl Future for CommitmentRequestProcessor {
                         warn!(?rpc_url, "websocket connection closed by server");
                         return Poll::Ready(InterruptReason::ConnectionClosed);
                     }
-                    Ok(_) => {} // ignore other message types
+                    Ok(Message::Binary(data)) => {
+                        warn!(
+                            ?rpc_url,
+                            bytes_len = data.len(),
+                            "received unexpected binary message"
+                        );
```
merklefruit: I suggest turning this down to debug! or trace! level, as it would be great to keep warnings for stuff that we need to take action or be careful of. same for Message::Frame

thedevbirb: Agreed

estensen: https://github.com/chainbound/bolt/pull/680/commits/6e208d8f09d9c8ea5a9e9ea2699da953ad9773fd

merklefruit (approved):

thedevbirb (approved):

PR #681, opened 2025-01-15
chore(bolt-sidecar): test calculate_max_basefee()

estensen:
Add tests to verify current behavior

Is adding `1` to to round up something that's intentional?

https://github.com/chainbound/bolt/blob/f60a51b20c887207eb6207a9d2039401afdaa72f/bolt-sidecar/src/common/transactions.rs#L30-L31


estensen:
> Is adding `1` to to round up something that's intentional?

I think overestimating the basefee is fine since we just wanna make sure we don't accept a lower fee than we should have.

thedevbirb:
I think adding 1 is used similarly to a `div_ceil`, and to ensure 0 is not returned. As such, a combination of `div_cel` and `max(value, 1)` is probably slightly better.

base fee should never be zero: https://github.com/ethereum/go-ethereum/blob/509a64ffb9405942396276ae111d06f9bded9221/consensus/misc/eip1559/eip1559.go#L74-L83

mempirate (approved):

PR #682, opened 2025-01-15
chore(bolt-boost): bump commit-boost-client to v0.5.0

estensen:
Now that there's a release with the alloy version we need use it instead of pinning commit


estensen:
Breaking change in https://github.com/Commit-Boost/commit-boost-client/pull/212

thedevbirb (approved):
Tested in devnet and works!

mempirate (approved):

Issue #673, opened 2025-01-13
Global sidecar state for safe shutdown


mempirate:
## Context
Currently, there's no way to gracefully shut down the sidecar without risking a commitment fault. We should have a global, cloneable state watcher with the following states:
- `Available`
- `PendingShutdown`
- `Shutdown`

Upon a first interrupt (CTRL+C), the state should be set to `PendingShutdown` (should be event based, like a [watch channel](https://docs.rs/tokio/1.43.0/tokio/sync/watch/index.html)). This event will be caught by all actor tasks, and will kick off a process that checks whether we have any upcoming proposer duties with commitments. If we do, this process will wait until that slot has passed, and only then fire a `Shutdown` event, which will result in the actual shutting down of all actors and finally the process itself.

The API should ALWAYS reflect the state in any API responses ([503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) if `PendingShutdown`). This ensures we can actually shut down without adding in new commitments.

So we should have a bi-directional communication channel:
- top level interrupt handler fires an event to all receivers
- receivers start cleaning up, and send an ack back up the channel when done
- only shut down when all acks are received

Note that this would be slightly less important with a persistent fallback block cache, but still important.

We should also have a way to force exit (i.e. press ctrl+c 3 times or something)

thedevbirb:
Slightly related: https://github.com/chainbound/bolt/pull/552#discussion_r1883590282 (cancellation tokens)

PR #683, opened 2025-01-15
fix(sidecar): ping messages were not sent

thedevbirb:
This PR fixes the ping of the `CommitmentRequestProcessor`. In essence, pings were sent only if other events woke up the task because the receiver is not using the asynchronous context.
Moreover, a broadcast receiver doesn't track individual tasks or their wakers. As such, I've opted to completely remove the `PingTicker` and use an interval instead spawned per task, that once ticked sends ping messages.

### DIFF
--- a/bolt-sidecar/src/api/commitments/firewall/receiver.rs
+++ b/bolt-sidecar/src/api/commitments/firewall/receiver.rs
@@ -25,10 +25,6 @@ use super::{
     processor::{CommitmentRequestProcessor, InterruptReason, ProcessorState},
 };
 
-/// The maximum number of retries to attempt when reconnecting to a websocket server.
-/// Try indefinitely.
-const MAX_RETRIES: usize = usize::MAX;
-
 /// The maximum messages size to receive via websocket connection, in bits, set to 32MiB.
 ///
 /// It is enough to account for a commitment request with 6 blobs and the largest
@@ -124,7 +120,7 @@ impl CommitmentsReceiver {
 
             tokio::spawn(async move {
                 retry_with_backoff_if(
-                    MAX_RETRIES,
+                    None,
                     Some(retry_config),
                     // NOTE: this needs to be a closure because it must be re-called upon failure.
                     // As such we also need to clone the inputs again.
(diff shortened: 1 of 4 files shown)


merklefruit:
Have you considered moving from sync::broadcast to sync::watch channels? 
Perhaps useless for the ping one since it needs to send repeated messages but can be useful to fix the shutdown one.

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -28,14 +25,9 @@ use super::{
...
-/// The interval at which to send ping messages from connected clients.
-#[cfg(test)]
-const PING_INTERVAL: Duration = Duration::from_secs(4);
-#[cfg(not(test))]
-const PING_INTERVAL: Duration = Duration::from_secs(30);
-
 /// The maximum number of retries to attempt when reconnecting to a websocket server.
-const MAX_RETRIES: usize = 1000;
+/// Try indefinitely.
+const MAX_RETRIES: usize = usize::MAX;
```
merklefruit: what's the point of having this vs an infinite loop?

thedevbirb: You still have a exponential backoff with jitter, no?

thedevbirb:
> Have you considered moving from sync::broadcast to sync::watch channels? Perhaps useless for the ping one since it needs to send repeated messages but can be useful to fix the shutdown one.

Added a workaround with https://github.com/chainbound/bolt/pull/683/commits/987410d6cb61be7a15e84ab6b4b1f2cb37e18095. https://github.com/chainbound/bolt/issues/673 would be a generalized solution.

bolt-sidecar/src/api/commitments/firewall/receiver.rs
merklefruit: Wdyt of using an Option<u64> for max_retries in retry_with_backoff_if ? 
then if the option is None, you skip the `.take(max_retries)` on the ExponentialBackoff stream.

just a thought

thedevbirb: I like it, thanks!

thedevbirb: https://github.com/chainbound/bolt/pull/683/commits/add06f1b47aecacfad38743cb04a4d69e3a76da8

merklefruit: Great! thanks!

merklefruit (approved):
LGTM

mempirate (approved):
Good stuff, the peculiarities of async rust init

PR #685, opened 2025-01-16
chore(sidecar, holesky): update example environment variables files

thedevbirb:
Updates such files with the firewall delegation options and removes the stale `BOLT_SIDECAR_COMMITMENT_PRIVATE_KEY`

### DIFF
--- a/bolt-sidecar/.env.example
+++ b/bolt-sidecar/.env.example
@@ -2,7 +2,15 @@
 
 # Port to listen on for incoming JSON-RPC requests of the Commitments API. This
 # port should be open on your firewall in order to receive external requests!
-BOLT_SIDECAR_PORT=8017
+BOLT_SIDECAR_PORT=
+# Comma-separated list of allowed RPC addresses to subscribe via websocket to receive
+# incoming commitments requests.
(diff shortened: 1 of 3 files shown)


bolt-sidecar/.env.example
```diff
@@ -2,7 +2,15 @@
...
 # port should be open on your firewall in order to receive external requests!
-BOLT_SIDECAR_PORT=8017
+BOLT_SIDECAR_PORT=
+# Comma-separated list of allowed RPC addresses to subscribe via websocket to receive
+# incoming commitments requests.
+# This is incompatible with the `BOLT_SIDECAR_RPC_URL` option.
+BOLT_SIDECAR_FIREWALL_RPCS="ws://my_rpc.io/api/v1/firewall_stream"
+# Secret ECDSA key to sign commitment messages with. The public key associated
+# to it must be then used when registering the operator in the `BoltManager`
+# contract
```
merklefruit: heads up that `BoltManager` is going to be deprecated after the hybrid registry release. 
I suggest either mentioning the `OperatorsRegistry` or no contract name at all

thedevbirb: fixed both reviews, lmk!

merklefruit (approved):

PR #691, opened 2025-01-17
fix(holesky): cb config + bump images for release

thedevbirb:
This PR fixes configuration files for running commit_boost with the new 0.4.0 release.
* bump docker images
* update docker compose file (align it to what the commit_boost CLI would output)
* update various infra files such as env.example and preset.env.example

PR #696, opened 2025-01-17
fix(holesky): Align CB config and env files

merklefruit:
Stacked on #691 

I had to remove the `depends_on` flag because the `cb_signer` container doesn't spin up correctly due to some config parsing error. Until that is fixed, we MUST not add this back as it will prevent the sidecar process from starting.

* add BOLT_SIDECAR_OPERATOR_PRIVATE_KEY required option in bolt-sidecar.env
* Align other missing/outdated vars from .env.example in bolt-sidecar.env
* Update formatting of .env.example for readibility

### DIFF
--- a/bolt-sidecar/.env.example
+++ b/bolt-sidecar/.env.example
@@ -1,57 +1,71 @@
-# Ethereum Node Connections + PBS URLs
+# --- Ethereum Node Connections + PBS URLs ---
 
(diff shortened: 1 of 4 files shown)


thedevbirb (approved):

estensen (approved):

mempirate (approved):
LGTM!

PR #700, opened 2025-01-17
chore: group dependency updates in one PR

thedevbirb:
So we don't get flooded. Reference: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--


merklefruit (approved):

mempirate:
If this doesn't work well we can try to use groups

mempirate (approved):

PR #713, opened 2025-01-20
chore(deps): group breaking and pre v1

estensen:
Pre v1 are breaking and should be grouped together with breaking changes, not together with non-breaking like currently in https://github.com/chainbound/bolt/pull/710


estensen:
`ignore` cannot be under `groups` and it's not allowed to have multiple configs targeting the same directory

PR #721, opened 2025-01-22
chore(cli, justfile): send preconfs also via `eth_sendRawTransaction`

thedevbirb:
Introduces a `--raw` flag inside the `send` command of `bolt` CLI to use `eth_sendRawTransaction` instead of `bolt_requestInclusion`. 

Updates the related Just commands to use this functionality.


estensen (approved):
🐶

merklefruit (approved):

PR #732, opened 2025-01-23
feat(sidecar): verify validator public key in bolt_metadata endpoint

### LINKED ISSUE #730: Accept validator pubkey as param in `bolt_metadata`
Firewall delegation, implemented in #552, introduced the ability for an operator opted into bolt to specify an entity (an RPC) responsible for filtering and forwarding inclusion preconfirmation requests to its sidecar, without the necessity to expose public endpoints.  The connection between the two is done via a websocket connection, authenticated with a JWT that contains a claim signed by the operator itself.

We first assumed that an operator would run its sidecar with all its associated validator public keys, but this is a wrong assumption. For high-availability setups, it makes sense that validator public keys are distributed on different machines and periodically rotating, thus many websocket connections should be opened by the same operator.

When new proposer duties are available, the bolt RPC currently calls the `bolt_metadata` endpoint to check whether the sidecar associated to a validator is online. By accepting this parameter, the sidecar can respond whether a validator is available on a certain machine or not, allowing the RPC to forward subsequent request to the correct websocket connection.

thedevbirb:
Closes #730 

Tested on devnet with bolt RPC on latest code that allows multiple websocket connections by the same operator

### DIFF
--- a/bolt-sidecar/src/api/commitments/firewall/processor.rs
+++ b/bolt-sidecar/src/api/commitments/firewall/processor.rs
@@ -1,3 +1,4 @@
+use ethereum_consensus::crypto::PublicKey;
 use futures::{
     pin_mut,
     stream::{FuturesUnordered, SplitSink, SplitStream},
@@ -5,7 +6,7 @@ use futures::{
 };
 use serde::Serialize;
 use serde_json::{json, Value};
-use std::time::Duration;
+use std::{collections::HashSet, time::Duration};
 use std::{collections::VecDeque, future::Future, pin::Pin, task::Poll};
 use tokio::{
     net::TcpStream,
@@ -82,16 +83,18 @@ impl Future for PendingCommitmentResponse {
 }
 
 /// The internal state of the [CommitmentRequestProcessor].
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct ProcessorState {
     /// The running limits of the sidecar.
     limits: LimitsOpts,
+    /// The available validator public keys in the sidecar.
+    available_validators: HashSet<PublicKey>,
 }
 
 impl ProcessorState {
     /// Creates a new instance of the [ProcessorState].
-    pub fn new(limits: LimitsOpts) -> Self {
-        Self { limits }
+    pub fn new(limits: LimitsOpts, available_validators: HashSet<PublicKey>) -> Self {
+        Self { limits, available_validators }
     }
 }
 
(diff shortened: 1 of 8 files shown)


bolt-sidecar/src/api/commitments/firewall/processor.rs
```diff
@@ -322,12 +325,37 @@ impl CommitmentRequestProcessor {
...
+                let public_key = request
+                    .params
+                    .first()
+                    .and_then(|p| p.as_str())
+                    .and_then(|p| hex::decode(p).ok())
+                    .and_then(|p| PublicKey::try_from(p.as_slice()).ok());
+
+                let response: JsonRpcResponse = 'v: {
+                    let Some(public_key) = public_key else {
+                        break 'v JsonRpcErrorResponse::new(
```
merklefruit: Something about this syntax irritates me for some reason, I cannot explain it 😁!
What's the benefit of this vs:

```rust
let response: JsonRpcResponse = if let Some(public_key) = public_key {
    if self.state.available_validators.contains(&public_key) {
        let metadata = MetadataResponse {
            limits: self.state.limits,
            version: BOLT_SIDECAR_VERSION.to_string(),
        };
        JsonRpcSuccessResponse::new(json!(metadata)).into()
    } else {
        JsonRpcErrorResponse::new(CommitmentError::ValidatorNotAvailable.into()).into()
    }
} else {
    JsonRpcErrorResponse::new(
        CommitmentError::InvalidParams("missing or invalid validator public key".into()).into(),
    )
    .into()
};
```

to me this version seems more readable and less nested but maybe I am missing something, what do you think?

thedevbirb: Ahah no worries I'm experimenting with it and see where it can be useful. The idea is that sometimes this can help you reduce indentation by creating something similar to a closure where you can early return, but without needing to provide all the parameters to the closure. It can be very handy.

In this specific scenario you're right, at first it was convenient but after some changes there is no particular benefit and the suggestion you've provided is indeed more readable.

Will address it soon!

bolt-sidecar/src/api/commitments/firewall/receiver.rs
```diff
@@ -137,6 +147,7 @@ impl CommitmentsReceiver {
 
                         let api_events_tx = api_events_tx.clone();
                         let shutdown_rx = shutdown_rx.clone();
+                        let state = state.clone();
```
merklefruit, line 150: what is this for?

bolt-sidecar/src/api/commitments/spec.rs
```diff
@@ -93,6 +97,9 @@ impl From<CommitmentError> for JsonRpcError {
             CommitmentError::InvalidParams(err) => Self::new(-32602, err.to_string()),
             CommitmentError::Internal => Self::new(-32603, err.to_string()),
             CommitmentError::RejectedJson(err) => Self::new(-32604, err.to_string()),
+            CommitmentError::ValidatorNotAvailable(ref pk) => {
+                Self::new(600, err.to_string()).with_data(json!(pk))
```
merklefruit, line 101: where does 600 come from? is it our custom spec?

thedevbirb: Yes. I know that it's not yet public but I thought it would be a good idea to start using it

merklefruit: No problem, I was just curious

merklefruit (approved):

PR #739, opened 2025-01-25
Feat: Make `bolt-sidecar` package a Cargo workspace

### LINKED ISSUE #706: Make `bolt-sidecar` package a Cargo workspace
Currently the `bolt-sidecar` is one big crate. As the project is becoming more and more mature there are some components (like the primitives) that should become separate crates so that can be used by other projects as dependency.

For a guideline, every folder under `bolt-sidecar/src` could become a separate crate.

0ex-d:
Closes #706


0ex-d:
Lots of dependencies across, duplicate first -> break things -> delete `botl-sidecar/src` -> fix -> test 🚀.

PR #746, opened 2025-01-28
test(sidecar): add basefee overflow test

mempirate:
The explicit error will make sure we don't get any overflows, unless basefee is ridiculously high (didn't calculate exactly, but for `block_diff = 64`, `current` would need to be way higher than 1 eth.

### DIFF
--- a/bolt-sidecar/src/state/execution.rs
+++ b/bolt-sidecar/src/state/execution.rs
@@ -74,6 +74,9 @@ pub enum ValidationError {
     /// The maximum commitments have been reached for the slot.
     #[error("Already requested a preconfirmation for slot {0}. Slot must be >= {0}")]
     SlotTooLow(u64),
+    /// The slot is too high (> current_slot + 64)
+    #[error("Target slot {0} is too high. Max target slot: {1}")]
+    SlotTooHigh(u64, u64),
     /// The maximum committed gas has been reached for the slot.
     #[error("Max committed gas reached for slot {0}: {1}")]
     MaxCommittedGasReachedForSlot(u64, u64),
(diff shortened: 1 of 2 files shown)


bolt-sidecar/src/state/execution.rs
```diff
@@ -291,6 +295,11 @@ impl<C: StateFetcher> ExecutionState<C> {
         // Check if the max_fee_per_gas would cover the maximum possible basefee.
         let slot_diff = target_slot.saturating_sub(self.slot);
 
+        // Verify max slot diff
+        if slot_diff > 64 {
+            return Err(ValidationError::SlotTooHigh(target_slot, self.slot + 64));
+        }
+
```
thedevbirb: While I support this change, the fact that we need it is weird.

If you take a look below:
https://github.com/chainbound/bolt/blob/f553103879ce6aec5a7ec1a06d7f14bc9b9c5746/bolt-sidecar/src/state/consensus.rs#L114-L122

you can see that when doing consensus checks we already verify that the slot of the inclusion request is in some bounds, so the a `slot_diff` greater than `64` should not happen because it should be larger then `furthest_slot`. Note that we do consensus checks _before_ execution checks

Can you please double check if there is another hidden error there? Happy to debug it together if needed

mempirate: Ah you're right, I missed this! Gonna remove the check.

merklefruit: note that this check will be skipped if the user has set --unsafe-skip-consensus-checks when running the sidecar. Therefore it might still be useful to have it (although in prod environment it won't make a difference)

mempirate:
Closing in favor of #681

PR #748, opened 2025-01-29
fix(guides): don't use quotes for single-word extradata

merklefruit:
An operator incurred into this issue and registered twice for this reason

### DIFF
--- a/guides/mainnet-genesis/README.md
+++ b/guides/mainnet-genesis/README.md
@@ -100,7 +100,8 @@ where:
 - `<your-rpc-url>` is the URL of the Ethereum RPC node you are using
 - `<your-operator-private-key>` is the private key of the operator account you are using
 - `<your-extra-data-string>` is any extra string you want to include in the registration,
-  such as your operator name, website or a custom identifier
+  such as your operator name, website or a custom identifier. If you use a single word, please
+  avoid using quotes (e.g. `my-operator` instead of `"my-operator"`).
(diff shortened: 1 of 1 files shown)


merklefruit:
## 📚 $\text{Stack Overview}$

Pulls submitted in this stack:
* #748 👈
* `unstable`

_This comment was automatically generated by [`st`](https://github.com/clabby/st)._

guides/mainnet-genesis/README.md
```diff
@@ -100,7 +100,8 @@ where:
 - `<your-rpc-url>` is the URL of the Ethereum RPC node you are using
 - `<your-operator-private-key>` is the private key of the operator account you are using
 - `<your-extra-data-string>` is any extra string you want to include in the registration,
-  such as your operator name, website or a custom identifier
+  such as your operator name, website or a custom identifier. If you use a single word, please
+  avoid using quotes (e.g. `my-operator` instead of `"my-operator"`).
```
thedevbirb, line 104: Some questions:
- can't we strip away `"` on the `bolt` CLI instead?
- I think "word" might be a little ambiguous since I don't consider a URL a word. Maybe you were referring about the presence of whitespaces
- can you make more examples of what is the on-chain result based on the input?

merklefruit: here is the example: https://etherscan.io/address/0x630869F51C012C797FEb3D9006F4280587C78b3f#readProxyContract#F6

PR #752, opened 2025-01-29
chore: remove all reth types

merklefruit:
* We can get away with `alloy::consensus::PooledTransaction` instead of `reth_primitives::TransactionSigned` 
* Similarly, `BlockBody, SealedBlock, SealedHeader` from reth can all be avoided with `alloy::consensus::Block`!

We don't have any reth types left after this. Upstack PR #753 removes them from Cargo.toml!

PR #753, opened 2025-01-29
chore(sidecar): rm unused dependencies

merklefruit:
We had some unused dependencies in bolt_sidecar.
Added a library attribute to warn us in the future.

Dependencies removed: 

* tokio-stream
* ssz_rs
* ethereum_ssz
* reth_primitives
* reth_primitives_traits
* bytes

### DIFF
--- a/bolt-sidecar/src/lib.rs
+++ b/bolt-sidecar/src/lib.rs
@@ -2,6 +2,11 @@
 #![warn(missing_debug_implementations, missing_docs, rustdoc::all)]
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+
+// Manually silence "rustls" crate being unused
+#[allow(unused_extern_crates)]
+extern crate rustls;
 
 /// All APIs in use by the sidecar.
 pub mod api;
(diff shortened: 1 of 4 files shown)


merklefruit:
## 📚 $\text{Stack Overview}$

Pulls submitted in this stack:
* #753 👈
* #752
* `unstable`

_This comment was automatically generated by [`st`](https://github.com/clabby/st)._

bolt-sidecar/src/lib.rs
```diff
@@ -2,6 +2,11 @@
 #![warn(missing_debug_implementations, missing_docs, rustdoc::all)]
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+
+// Manually silence "rustls" crate being unused
+#[allow(unused_extern_crates)]
+extern crate rustls;
```
thedevbirb, line 9: Why is this needed?

thedevbirb (approved):

bolt-sidecar/src/lib.rs
merklefruit: because rustls isn't used in the lib/ code, so the "unused crate" warning is triggered, but we do use it in bin/sidecar.rs, which isn't checked by this particular rule.

thedevbirb: Got it, related to #706

PR #756, opened 2025-01-31
Bolt-cli: Add dry-run flag for developmental/testing

### LINKED ISSUE #416: feat(cli): add `--dry-run` flag for important commands
We should add a `--dry-run` flag to all CLI commands that require an RPC connection.

When this is specified, we should use an Anvil chain fork as blockchain provider. 
This will be very handy for people looking to register validators or opt-in to various contracts.

0ex-d:
### Improvements:
- Add -d, --dry-run flag for bolt-cli: When provided commands will "simulate" actual behavior, for instance when making RPC calls or executing smart contracts without broadcasting on net. 
 - Spin Anvil instance for use cases like testing (we can get rid of explicit declaration in test):
 https://github.com/chainbound/bolt/blob/6342fb2ba6dfafbdf518f0c4d32295fca9e8188c/bolt-cli/src/commands/validators.rs#L165-L171

Closes #416

### DIFF
--- a/bolt-cli/src/cli.rs
+++ b/bolt-cli/src/cli.rs
@@ -187,6 +187,11 @@ pub enum ValidatorsSubcommand {
         /// The private key to sign the transactions with.
         #[clap(long, env = "ADMIN_PRIVATE_KEY")]
         admin_private_key: B256,
+
+        /// Run the command in "dry run" mode, run all steps without broadcast.
+        /// Useful for testing and debugging purposes.
+        #[clap(short, long, env = "DRY_RUN", default_value = "false")]
+        dry_run: bool,
     },
     /// Check the status of a validator (batch).
     Status {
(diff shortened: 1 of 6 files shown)


0ex-d:
Hi @merklefruit, just checking in on this PR. Let me know if any changes are needed. Looking forward to your feedback. Thanks!

bolt-cli/src/cli.rs
```diff
@@ -304,6 +325,10 @@ pub enum EigenLayerSubcommand {
         /// The URL of the RPC to read data from
         #[clap(long, env = "RPC_URL")]
         rpc_url: Url,
+        /// Run the command in "dry run" mode, run all steps without broadcast.
+        /// Useful for testing and debugging purposes.
+        #[clap(short, long, env = "DRY_RUN", default_value = "false")]
+        dry_run: bool,
```
merklefruit: listing strategies is read-only, as such we don't need to simulate this one

bolt-cli/src/cli.rs
```diff
@@ -365,6 +402,10 @@ pub enum SymbioticSubcommand {
...
```
merklefruit: same here for ListVaults

merklefruit (changes requested):
Great work here! Really sorry about the delay. Thanks a lot for the contribution!

I left a couple nits :)

merklefruit (approved):
lgtm, thanks!

PR #769, opened 2025-02-07
fix: typos in documentation files

maximevtush:


estensen:
Hi @maximevtush, thank you for you're interest, but we currently don't accept contributions for typos.  
https://github.com/chainbound/bolt/blob/unstable/CONTRIBUTING.md#exclusions

PR #776, opened 2025-02-17
chore:Bolt-cli:Update Readme usage section new cmds

0ex-d:
Some recently added commands were missing from doc


estensen (approved):
Ty for this

